PR #22, opened 2026-04-13
feat: Downstream llm-d

albertoperdomo2:
## Summary by CodeRabbit

* **New Features**
  * Added comprehensive runtime orchestration for LLM D deployment on Kubernetes/OpenShift clusters
  * Implemented model caching for HuggingFace and OCI registry models with PVC-based storage
  * Added multiple scheduler profiles (`approximate`, `precise`) for inference optimization
  * Introduced new runtime presets (`smoke`, `benchmark-short`) for different workload configurations
  * Added preparation, testing, and cleanup toolbox workflows
  * Implemented smoke testing and GuideLLM benchmarking capabilities

* **Documentation**
  * Updated project README with llm_d-specific orchestration scope


config/llm_d/models.yaml
```diff
@@ -0,0 +1,25 @@
+models:
```
kpouget: as discussed offline, we'll need a mechanism that supports multiple configuration files

this will have to go to `projects/llm_d/orchestration/config.d`

projects/llm_d/orchestration/ci.py
```diff
@@ -1,48 +1,37 @@
...
 
 import sys
-import subprocess
-import time
 from pathlib import Path
-import types
+
+FORGE_HOME = Path(__file__).resolve().parents[3]
+if str(FORGE_HOME) not in sys.path:
+    sys.path.insert(0, str(FORGE_HOME))
```
kpouget: `env.FORGE_HOME`

projects/llm_d/orchestration/prepare_llmd.py
```diff
@@ -1,15 +1,428 @@
...
+    prepare_nfd(config)
+    prepare_gpu_operator(config)
+    prepare_rhoai_operator(config)
+    apply_datasciencecluster(config)
+    wait_for_datasciencecluster_ready(config)
+    ensure_required_crds(config.platform["rhoai"]["required_crds_after_dsc"], config)
+    ensure_gateway(config)
+    ensure_test_namespace(config)
+    verify_gpu_nodes(config)
+    capture_prepare_state(config)
```
kpouget: this will have to become toolbox commands

global config is at projects.core.library.config --> config.project.get_config

projects/core/library/config.py
```diff
@@ -335,20 +334,73 @@ def filter_out_used_overrides(self):
...
 
-    logger.info(f"Copying the configuration from {config_file_src} to the artifact dir ...")
-    shutil.copyfile(config_file_src, config_path_final)
+    logger.info(
+        f"Consolidating the configuration from {config_file_src} "
+        f"and {config_dir_src} to the artifact dir ..."
+    )
+    with open(config_path_final, "w") as config_f:
+        yaml.safe_dump(
+            _load_project_config(orchestration_dir),
```
kpouget:
```suggestion
    config_dict = _load_project_config(orchestration_dir)
    
    with open(config_path_final, "w") as config_f:
        yaml.safe_dump(
            config_dict,
```
`_load_project_config` is an important operation, better not hide it in the middle of the yaml file dump

projects/core/library/config.py
```diff
@@ -335,20 +334,73 @@ def filter_out_used_overrides(self):
...
+            default_flow_style=False,
+            sort_keys=False,
+        )
 
     return config_path_final, config_file_src
 
 
+def _load_project_config(orchestration_dir):
+    config_file_src = orchestration_dir / "config.yaml"
+    config_dir_src = orchestration_dir / "config.d"
```
kpouget: these two should come from the parameters, `__get_config_path` already define them

projects/core/library/config.py
```diff
@@ -335,20 +334,73 @@ def filter_out_used_overrides(self):
...
+    if not config_dir_src.exists():
+        return config
+
+    for chunk_file in sorted(config_dir_src.glob("*.yaml")):
+        with open(chunk_file) as chunk_f:
+            chunk_value = yaml.safe_load(chunk_f)
+
+        key = chunk_file.stem
+        if key in config:
+            config[key] = _merge_config_values(config[key], chunk_value)
```
kpouget: I think we better go this way:
```suggestion
        if key in config:
            raise ValueException(f"Key {key} defined in the main config file and in {chunk_file}")
```
merging is unpredictable, in the the sense that you don't expect to have half of the field defined in one place, half in another side

projects/core/library/config.py
```diff
@@ -335,20 +334,73 @@ def filter_out_used_overrides(self):
...
+        else:
+            config[key] = chunk_value
+
+    return config
+
+
+def _merge_config_values(base_value, chunk_value):
```
kpouget: will be removed (cf comment above)

projects/core/library/config.py
```diff
@@ -5,7 +5,6 @@
 import os
 import pathlib
 import re
-import shutil
```
kpouget: can you open a PR for the modification of this file?

projects/core/dsl/log.py
```diff
@@ -117,6 +117,6 @@ def _get_toolbox_function_name(filename):
     """Extract toolbox function name from file path (parent directory name)"""
     filename_path = Path(filename)
 
-    # For paths like projects/llm_d/toolbox/capture_isvc_state/main.py
-    # Return the parent directory name: capture_isvc_state
+    # For paths like projects/llm_d/toolbox/capture_llmisvc_state/main.py
+    # Return the parent directory name: capture_llmisvc_state
```
kpouget: can you remove these lines of comments, and merge this `_get_toolbox_function_name` with the one of `runtime.py`?

projects/llm_d/orchestration/config.d/platform.yaml
```diff
@@ -0,0 +1,85 @@
+cluster:
+  minimum_openshift_version: "4.19.9"
+  namespace_name: forge-llm-d
+  namespace_prefix: llm-d
+  namespace_max_length: 63
```
kpouget: usually when you start sharing prefixes, it's a sign you want to create a new dict:

```suggestion
  namespace:
      name: forge-llm-d
      prefix: llm-d
      max_length: 63
```

projects/llm_d/orchestration/config.d/platform.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - feature.node.kubernetes.io/pci-10de.present
+    - feature.node.kubernetes.io/pci-0302_10de.present
+    - feature.node.kubernetes.io/pci-0300_10de.present
+
+operators:
+  - display_name: OpenShift Cert Manager
+    package: openshift-cert-manager-operator
+    namespace: openshift-cert-manager-operator
+    channel: stable-v1.18
+    source: redhat-operators
```
kpouget, line 19: can you make that a dict? 

```suggestion
operators:
  openshift-cert-manager-operator:
    display_name: OpenShift Cert Manager
    namespace: openshift-cert-manager-operator
    channel: stable-v1.18
    source: redhat-operators
    enabled: true
```
dicts are way easier to control with JSONPATH
like `operators.openshift-cert-manager-operator.enabled: false`
which is safer than 
`like `operators[0].enabled: false`
and simpler than the equivalent search by `package=....`

projects/llm_d/orchestration/config.yaml
```diff
@@ -1,230 +1,3 @@
...
-      fail_test_on_regression: true
-      notification:
-        enabled: true
-        title: "llm-d CPT"
-  download:
-    mode: prefer_cache
-    url: null
+project:
+  name: llm_d
+  args: []
```
kpouget: (I'll make that go away, sooner than later)

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+from collections.abc import Iterable
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+from projects.core.library import config, env, run
+
+LOGGER = logging.getLogger(__name__)
```
kpouget: can you use `logger`, for consistency with the rest of the project? 
unless that's a common practice ? (it's a constant, but ...)

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+ORCHESTRATION_DIR = env.FORGE_HOME / "projects" / "llm_d" / "orchestration"
```
kpouget: you can use something like `Pathlib.Path(__file__).parent` for that

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    oci_registry_auth_secret_key: str | None
+
+    @property
+    def marker_path(self) -> str:
+        return f"/cache/{self.model_path}/{self.marker_filename}"
+
+
+def init() -> Path:
+    if not logging.getLogger().handlers:
+        logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
```
kpouget: this is now part of `env.init()`

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    env.init()
+    run.init()
+    ensure_artifact_directories(env.ARTIFACT_DIR)
+    return env.ARTIFACT_DIR
+
+
+def ensure_artifact_directories(artifact_dir: Path) -> None:
+    for relative in ("src", "artifacts", "artifacts/results"):
+        (artifact_dir / relative).mkdir(parents=True, exist_ok=True)
+
```
kpouget:
```suggestion
    ensure_artifact_subdirectories()
    return env.ARTIFACT_DIR


def ensure_artifact_subdirectories() -> None:
    for relative in ("src", "artifacts", "artifacts/results"):
        (env.ARTIFACT_DIR / relative).mkdir(parents=True, exist_ok=True)
```
I think it's easier to grasp this way

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+
+def load_run_configuration(
+    *, cwd: Path | None = None, artifact_dir: Path | None = None
+) -> ResolvedConfig:
+    cwd = cwd or Path.cwd()
+    if artifact_dir is not None:
+        os.environ["ARTIFACT_DIR"] = str(artifact_dir)
+    artifact_dir = init()
+    _reinitialize_project_config()
```
kpouget: this shouldn't be needed 🤔

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    overrides = parse_overrides(
+        os.environ.get("FORGE_CONFIG_OVERRIDES", ""),
+        allowed_keys=config.project.get_config("runtime.allowed_override_keys", []),
+    )
+
+    requested_preset = (
+        fournos_config.get("preset")
+        or os.environ.get("FORGE_PRESET")
+        or config.project.get_config("runtime.default_preset")
+    )
```
kpouget: this shouldn't be needed (the `os.environ.get` part)

one way to avoid this to have something like that in `cli.py`:
```
    if something:
        config.project.set_config("something", something)
        logger.info(f"Using something={something}")
```
this way the config is the only source of truth, and you can prepare it conveniently from the CLI
if the environ var is import, in the `cli.py` you do
```
if something := os.environ.get("SOMETHING"):
        config.project.set_config("something", something)
        logger.info(f"Using something={something}")  
```

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+
+    model_name = config.project.get_config("runtime.model_key")
+    model = copy.deepcopy(config.project.get_config(f"models.{model_name}"))
+
+    scheduler_profile_key = config.project.get_config("runtime.scheduler_profile_key")
+    scheduler_profile = None
+    if scheduler_profile_key != "default":
+        scheduler_profile = copy.deepcopy(
+            config.project.get_config(f"scheduler_profiles.{scheduler_profile_key}")
+        )
```
kpouget: could be interesting to turn this into
```suggestion
        scheduler_profile = config.project.get_config(f"scheduler_profiles.{scheduler_profile_key}", copy=True)
```
as it's a legit pattern to want to update the config. 
Or make it the default 🤔 the config fields won't be heavyweight anyway, and modifying the config directly should be prohibited (should be done via `set_config`)

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    default_namespace = platform_data["cluster"].get("namespace_name")
+    namespace = (
+        namespace_override
+        or default_namespace
+        or derive_namespace(
+            job_name,
+            platform_data["cluster"]["namespace_prefix"],
+            platform_data["cluster"]["namespace_max_length"],
+        )
+    )
```
kpouget: please add some logging to tell what's happening, I think at least here it's relevant

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+) -> ResolvedConfig:
+    cwd = cwd or Path.cwd()
+    if artifact_dir is not None:
+        os.environ["ARTIFACT_DIR"] = str(artifact_dir)
+    artifact_dir = init()
+    _reinitialize_project_config()
+
+    platform_data = copy.deepcopy(config.project.get_config("platform"))
+    model_cache = copy.deepcopy(config.project.get_config("model_cache"))
+    fournos_config = load_fournos_config(cwd)
```
kpouget: we'll find a way to make that transparent, via the config file.
We'll define a mapping to specify where in the config file the FOURNOS configuration should be stored

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    overrides = parse_overrides(
+        os.environ.get("FORGE_CONFIG_OVERRIDES", ""),
+        allowed_keys=config.project.get_config("runtime.allowed_override_keys", []),
+    )
+
+    requested_preset = (
+        fournos_config.get("preset")
+        or os.environ.get("FORGE_PRESET")
+        or config.project.get_config("runtime.default_preset")
```
kpouget: the config presets, as specified in the test launch, are applied as part of the `config.init()`, so no need to handle them manually

> `or config.project.get_config("runtime.default_preset")`
the "default" preset is the plain text config, before anything is applied
(if you always apply a default preset, then the plain text value is never used 🙃)
in the plain text config, I usually specify a simple lightweight smoke test

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    config.project = None
+    artifact_config = env.ARTIFACT_DIR / "config.yaml"
+    if artifact_config.exists():
+        artifact_config.unlink()
+
+    presets_applied = env.ARTIFACT_DIR / "presets_applied"
+    if presets_applied.exists():
+        presets_applied.unlink()
+
+    config.init(ORCHESTRATION_DIR)
```
kpouget: this shouldn't be needed, I'll need to understand why you had to use it (my I have my guess :)

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+
+
+def apply_requested_preset(requested_preset: str) -> None:
+    if not config.project.get_preset(requested_preset):
+        raise ValueError(f"Unknown llm_d preset: {requested_preset}")
+
+    config.project.apply_preset(requested_preset)
+
```
kpouget: already part of core/library/env.py

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    config_path = cwd / "fournos_config.yaml"
+    if not config_path.exists():
+        return {}
+
+    data = load_yaml(config_path)
+    if data is None:
+        return {}
+    if not isinstance(data, dict):
+        raise ValueError(f"Unexpected FOURNOS config type in {config_path}: {type(data)}")
+    return data
```
kpouget: not needed

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+
+
+def parse_overrides(raw: str, *, allowed_keys: Iterable[str]) -> dict[str, Any]:
+    if not raw or raw.strip() in {"", "null", "{}"}:
+        return {}
+
+    try:
+        data = json.loads(raw)
+    except json.JSONDecodeError as exc:
+        raise ValueError(f"FORGE_CONFIG_OVERRIDES is not valid JSON: {exc}") from exc
```
kpouget: Image

XD

we'll need to discuss the allowed_keys/unsupport, I'm not sure what's not supported 🤔

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with path.open("w", encoding="utf-8") as handle:
+        json.dump(payload, handle, indent=2, sort_keys=True)
+        handle.write("\n")
+
+
+def write_text(path: Path, content: str) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(content, encoding="utf-8")
+
```
kpouget: those could be part of a core.library package

projects/llm_d/runtime/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+        if result.stderr:
+            LOGGER.info("stderr:\n%s", result.stderr.rstrip())
+
+    if check and result.returncode != 0:
+        raise CommandError(
+            f"Command failed with exit code {result.returncode}: "
+            f"{' '.join(shlex.quote(arg) for arg in cmd)}"
+        )
+
+    return result
```
kpouget, line 158: there's already a wrapper for that in core.library.run --> run()

projects/llm_d/runtime/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+        check=check,
+        capture_output=capture_output,
+        input_text=input_text,
+        timeout_seconds=timeout_seconds,
+    )
+
+
+def apply_manifest(artifact_path: Path, manifest: dict[str, Any]) -> None:
+    write_yaml(artifact_path, manifest)
+    oc("apply", "-f", str(artifact_path))
```
kpouget, line 179: could be part of a project.core.library.oc package

kpouget: same for the functions below

projects/llm_d/runtime/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    normalized = stderr.lower()
+    if "error from server (notfound)" in normalized:
+        return True
+    if "no resources found" in normalized:
+        return True
+
+    return bool(re.search(r"\bnot found\b", normalized))
+
+
+def wait_until(
```
kpouget, line 237: this typically should be in a toolbox task, rather than in the orchestration layer

kpouget: same for the functions below

projects/llm_d/runtime/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+    operator_group = {
+        "apiVersion": "operators.coreos.com/v1",
+        "kind": "OperatorGroup",
+        "metadata": {"name": package, "namespace": namespace},
+        "spec": {"targetNamespaces": [namespace]},
+    }
+    oc("apply", "-f", "-", input_text=yaml.safe_dump(operator_group, sort_keys=False))
+
+
+def ensure_subscription(operator_spec: dict[str, Any]) -> None:
```
kpouget, line 334: those are typically functions/tasks that should be in a toolbox
basically, we'll have code in different places:
* the project specific orchestration --> not meant to be used by other projects
* the orchestration libraries --> meant to be used by other orchestrations
* the project toolbox DSL --> "APIs" meant to be used by this project and, perhaps by other projects
* the toolbox libraries --> helpers for writing toolbox tasks, compatible with the DSL

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+        if name:
+            return name
+    return None
+
+
+def render_datasciencecluster(config: ResolvedConfig) -> dict[str, Any]:
+    template_path = config.config_dir / config.platform["rhoai"]["datasciencecluster_template"]
+    manifest = load_yaml(template_path)
+    manifest["metadata"]["name"] = config.platform["rhoai"]["datasciencecluster_name"]
+    manifest["metadata"]["namespace"] = config.platform["rhoai"]["namespace"]
```
kpouget: I think the DSC is cluster wide ;-)

projects/llm_d/runtime/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+            "resources": {"requests": {"storage": spec.pvc_size}},
+        },
+    }
+    if spec.storage_class_name:
+        manifest["spec"]["storageClassName"] = spec.storage_class_name
+    return manifest
+
+
+def render_model_cache_job(config: ResolvedConfig, spec: ModelCacheSpec) -> dict[str, Any]:
+    common_env = [
```
kpouget, line 504: we'll need to move those to toolbox commands, 
and in particular avoid embedding code in code:
```
    if spec.source_scheme == "hf":
        command = """
set -euo pipefail
```
better have it in the toolbox directory, as a plain text file.
Same for the PVC spec

kpouget: wait I missed it, it's Python->bash->Python :D 
```
        command = """
set -euo pipefail
mkdir -p "${MODEL_TARGET_DIR}"
rm -rf "${MODEL_TARGET_DIR}"/*
python -m pip install --quiet --no-cache-dir 'huggingface_hub[hf_xet]'
python - <<'PY'
import os
from huggingface_hub import snapshot_download
```

projects/llm_d/orchestration/llmd_runtime.py
```diff
@@ -0,0 +1,1242 @@
...
+            "accessModes": ["ReadWriteOnce"],
+            "resources": {"requests": {"storage": config.benchmark["pvc_size"]}},
+        },
+    }
+
+
+def render_guidellm_job(config: ResolvedConfig, endpoint_url: str) -> dict[str, Any]:
+    if not config.benchmark:
+        raise ValueError("Benchmark configuration is not enabled for this preset")
+
```
kpouget: all of that will go into toolbox commands

I'll need to understand how the `ResolvedConfig` differ from the main config

projects/llm_d/orchestration/test_llmd.py
```diff
@@ -1,29 +1,12 @@
...
 
-from projects.core.library import config, env, run
-from projects.llm_d.toolbox.capture_isvc_state.main import run as capture_isvc_state
+from projects.llm_d.orchestration import llmd_runtime
+from projects.llm_d.toolbox.test.main import run as test_toolbox_run
 
-logger = logging.getLogger(__name__)
 
+def init() -> None:
+    llmd_runtime.init()
```
kpouget: `llmd_runtime.py` is a huuuge file! 
`test_llmd` and `prepare_llmd` are useless here, in the current form you could get rid of them and invoke directly `llmd_runtime`

many functions will go out of `llmd_runtime` anyway, 
we'll see if we can split it further into `prepare` and `test` code or not

projects/llm_d/toolbox/cleanup/main.py
```diff
@@ -0,0 +1,123 @@
...
+
+def run() -> int:
+    llmd_runtime.init()
+    config = llmd_runtime.load_run_configuration()
+    return run_cleanup(config)
+
+
+def run_cleanup(config: llmd_runtime.ResolvedConfig) -> int:
+    delete_run_leftovers(config)
+    return 0
```
kpouget: we'll rediscuss the toolbox together
you need @task annotations

projects/llm_d/toolbox/cleanup/main.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+        llmd_runtime.oc(
+            "delete",
+            "pod",
+            f"{benchmark_name}-copy",
+            "-n",
+            namespace,
+            "--ignore-not-found=true",
+            check=False,
+        )
```
kpouget: I'm not convinced by this form, I find this way more readable
```suggestion
    shell.run(f"oc delete job,pvc {benchmark_name} - n {namespace} --ignore-not-found=true", check=False)
    shell.run(f"oc delete pod {benchmark_name}-copy -n {namespace} --ignore-not-found=true", check=False)
```
and the shell injection isn't much of a worry in our context IMO

projects/llm_d/toolbox/prepare/main.py
```diff
@@ -0,0 +1,368 @@
...
+    prepare_rhoai_operator(config)
+    apply_datasciencecluster(config)
+    wait_for_datasciencecluster_ready(config)
+    ensure_required_crds(config.platform["rhoai"]["required_crds_after_dsc"], config)
+    ensure_gateway(config)
+    ensure_test_namespace(config)
+    cleanup_toolbox.delete_run_leftovers(config)
+    prepare_model_cache.run_prepare_model_cache(config)
+    verify_gpu_nodes(config)
+    capture_prepare_state(config)
```
kpouget: we'll see together how to rewrite the toolbox.

An important aspect of the toolbox is that it doesn't use the main config, 
this way it's strictly limited to its input parameters

tests/llm_d/test_runtime.py
```diff
@@ -5,7 +5,9 @@
 
 import pytest
 
-from projects.llm_d.orchestration import llmd_runtime
+from projects.llm_d.orchestration import ci as llmd_ci
+from projects.llm_d.orchestration import cli as llmd_cli
+from projects.llm_d.orchestration import llmd_runtime, phase_inputs
```
kpouget: @albertoperdomo2, can you move this to `projects/llm_d/tests`?

I would like to strive to have `projects/xxx` to be standalone, so that we can move it to another repo in the future if relevant

look for instance https://github.com/openshift-psap/fournos/tree/main/tests/forge/deploy
that's a standalone external test directory

kpouget:
thanks @albertoperdomo2 , let's merge this version, looks good overall, thanks for the effort 🙏🏻 

test artifacts: https://mlflow.apps.aperdomo-lab.ibm.rhperfscale.org/#/experiments/52/runs/e60d312b188e4266934a2973d1fdb977/artifacts

PR #59, opened 2026-05-13
[fournos] add fournos gitops

kpouget:


kpouget:
merging this to try the gitops deployment

PR #68, opened 2026-06-02
feat: add rhaiis project for KServe InferenceService benchmarking

MML-coder:
## Summary

Adds the `projects/rhaiis/` project for benchmarking AI inference engines on OpenShift using KServe InferenceService + GuideLLM.

### What's included

- **4 rhaiis toolbox commands**: `deploy_kserve_isvc`, `wait_isvc_ready`, `capture_isvc_state`, `cleanup_isvc`
- **Reuses canonical `run_guidellm_benchmark`** from `projects/guidellm/` (shared with llm_d) — no forked benchmark runner
- CLI with dry-run, model/workload selection, CLI overrides (`--rates`, `--max-seconds`, `--tensor-parallel`, etc.)
- Fournos integration: `resolve-fournos-config`, `export-artifacts`, vault support, prepare/cleanup
- 52 models from model_furnace + 4 workload profiles (profile1-4)
- All toolbox scripts use shared `core/dsl/utils` (oc(), oc_get_json(), write_text(), condition_status())
- Config-driven with presets for cluster-specific defaults

### Commits

1. `feat: add rhaiis project` — initial toolbox commands, CLI, config, presets
2. `feat: add Fournos integration` — resolve-fournos-config, export-artifacts, vault, prepare/cleanup, @entrypoint, split KServe templates
3. `config.yaml updates` — 52 models from model_furnace, profile1-4 workloads
4. `fix: benchmark timeout` — increase from 900s to 14400s (4 hours)
5. `refactor: shared utils` — migrate all toolbox scripts to core/dsl/utils
6. `refactor: canonical guidellm` — replace rhaiis fork with canonical runner, override retry for rhaiis timeout
7. `docs: CLI overrides` — document override flags and quick test examples

## Workflow

```
deploy_kserve_isvc → wait_isvc_ready → run_guidellm_benchmark → capture_isvc_state → cleanup_isvc
```

On failure at any step, `capture_isvc_state` and `cleanup_isvc` still run (try/finally).

### E2E test 1 — rhaiis fork (before canonical migration)
- **Cluster**: mehulvalidation (kube:admin)
- **Namespace**: kserve-e2e-perf
- **Model**: qwen3-0_6b (Qwen/Qwen3-0.6B), 1 GPU
- **Workload**: profile1 (rates 1,50,100,200,300 × 450s)
- **Command**:
  ```bash
  python3 -m projects.rhaiis.orchestration.cli test \
    --model qwen3-0_6b --workload profile1 \
    --namespace kserve-e2e-perf \
    --image-pull-secret npalaska-image-pull \
    --service-account-name sa
  ```
- **Result**: PASS — 1.4MB benchmarks.json extracted, all resources cleaned up

### E2E test 2 — canonical runner (after migration)
- **Cluster**: mehulvalidation (kube:admin)
- **Namespace**: kserve-e2e-perf
- **Model**: qwen3-0_6b, 1 GPU
- **Workload**: profile1 with CLI overrides `--rates 1,5 --max-seconds 60`
- **Command**:
  ```bash
  python3 -m projects.rhaiis.orchestration.cli test \
    --model qwen3-0_6b --workload profile1 \
    --namespace kserve-e2e-perf \
    --image-pull-secret npalaska-image-pull \
    --service-account-name sa \
    --rates 1,5 --max-seconds 60
  ```
- **Result**: PASS — 1.4MB benchmarks.json extracted, all resources cleaned up
- **Verified**: CLI overrides correctly override workload profile defaults

### Dry-run validation
- `--dry-run` flag prints resolved config (model, workload, vllm_args, rates, namespace) without deploying
- Verified config resolution for multiple models (qwen3-0_6b, llama-4-scout-fp8)

### Bugs found and fixed during testing
- **Copy pod cleanup ordering**: PVC stuck in `Terminating` because copy pod held a volume mount. Fixed by deleting copy pod before PVC.
- **Canonical retry limit**: `@retry(attempts=180)` = 30 min max, too short for profile1. Fixed by overriding retry config from rhaiis callers: `wait_guidellm_benchmark_task._retry_config["attempts"] = timeout // 10`.

## Test plan

- [x] Dry-run CLI passes
- [x] E2E on mehulvalidation with rhaiis fork (pre-migration)
- [x] E2E on mehulvalidation with canonical runner (post-migration)
- [x] CLI overrides (--rates, --max-seconds) work correctly
- [x] Cleanup runs on success and failure (try/finally)
- [x] ruff check + ruff format pass
- [ ] Test on AMD accelerator
- [ ] Fournos-driven execution (resolve-fournos-config → test)


projects/rhaiis/toolbox/deploy_kserve_isvc/templates/kserve.yaml.j2
```diff
@@ -0,0 +1,124 @@
...
+  volumes:
+    - name: shared-memory
+      emptyDir:
+        medium: Memory
+        sizeLimit: 8Gi
+{%- endif %}
+---
+apiVersion: serving.kserve.io/v1beta1
+kind: InferenceService
+metadata:
```
kpouget: please split that in two files

projects/rhaiis/toolbox/deploy_kserve_isvc/main.py
```diff
@@ -0,0 +1,84 @@
...
+@task
+def apply_kserve_manifest(args, context):
+    shell.run(f"oc apply -f {context.kserve_manifest_path}")
+    return f"Applied ServingRuntime + InferenceService {args.deployment_name}"
+
+
+main = toolbox.create_toolbox_main(run)
+
+if __name__ == "__main__":
+    main()
```
kpouget: please add the `@entrypoint` annotation to `run` and
```suggestion
if __name__ == "__main__":
    run.main()
```
IIRC

kpouget: same in all your toolbox scripts

projects/rhaiis/toolbox/run_guidellm_benchmark/main.py
```diff
@@ -0,0 +1,336 @@
...
+        "spec": {
+            "accessModes": ["ReadWriteOnce"],
+            "resources": {"requests": {"storage": pvc_size}},
+        },
+    }
+
+
+def _render_copy_pod(*, name: str, namespace: str, image: str,
+                     pvc_name: str, node_name: str | None = None) -> dict:
+    pod = {
```
kpouget: please use a YAML template for this, it's more readable that inline python/json

projects/rhaiis/toolbox/run_guidellm_benchmark/main.py
```diff
@@ -0,0 +1,336 @@
...
+                }
+            ],
+        },
+    }
+    if node_name:
+        pod["spec"]["nodeName"] = node_name
+    return pod
+
+
+def run(
```
kpouget: please move this to the top, or move the top helps to a `utils.py` function

projects/rhaiis/toolbox/run_guidellm_benchmark/main.py
```diff
@@ -0,0 +1,336 @@
...
+            c.get("type") == "Ready" and c.get("status") == "True"
+            for c in conditions
+        )
+
+    _wait_until(
+        f"copy pod {context.copy_pod_name}",
+        timeout_seconds=120,
+        interval_seconds=5,
+        predicate=_helper_ready,
+    )
```
kpouget: we'll need to rewrite that to use the `@retry`  annotation in the task

kpouget:
wrong project, sorry !

projects/rhaiis/orchestration/ci.py
```diff
@@ -0,0 +1,82 @@
...
+
+def list_vaults() -> list[str]:
+    test_rhaiis.init()
+    return config.project.get_config("vaults")
+
+
+def resolve_hardware_request(hardware_spec: dict) -> dict:
+    test_rhaiis.init()
+
+    model_key = config.project.get_config("tests.rhaiis.model_key")
```
albertoperdomo2: `config.project.get_config("")` is called enough times throughout the project that it deserves its own abstraction like [llm_d/orchestration/runtime_config.py](https://github.com/openshift-psap/forge/blob/main/projects/llm_d/orchestration/runtime_config.py), it will make things easier and avoid config resolution being scattered everywhere. See for example [this](https://github.com/openshift-psap/forge/blob/main/projects/llm_d/orchestration/test_phase.py#L30).

albertoperdomo2: I would move `_merge_vllm_args`, `_merge_env_vars`, `_build_guidellm_args`, `_split_image_tag`, and all `config.project.get_config()` resolution intoit, and then both CLI and CI can call the same accessors.

projects/rhaiis/orchestration/test_rhaiis.py
```diff
@@ -0,0 +1,171 @@
...
+logger = logging.getLogger(__name__)
+
+
+def init():
+    env.init()
+    run.init()
+    config.init(pathlib.Path(__file__).parent)
+
+
+def _merge_vllm_args(
```
albertoperdomo2: All of this methods: `_merge_vllm_args`, `_merge_env_vars`, `_build_guidellm_args`, `_split_image_tag` are duplicated in `test_rhaiis.py` and `ci.py`. I suggest extracting a single `test_phase.py`, so the CLI test command should resolve config and then call `test_phase.run()`.

projects/rhaiis/toolbox/capture_isvc_state/main.py
albertoperdomo2, line 1: In the `llm_d` project, we directly import and call the kserve toolbox:

```
from projects.kserve.toolbox.capture_llmisvc_state import main as capture_llmisvc_state
```

This file implements an almost identical toolbox command, which can be problematic e.g. when kserve's capture toolbox gets a bug fix or new capture task, RHAIIS won't get it. I suggest either parameterizing the kserve capture toolbox (resource type, label selector as arguments), or create a thin wrapper that calls the kserve toolbox and adds the events-capture task.

projects/rhaiis/orchestration/config.yaml
albertoperdomo2, line 1: I think the right approach would be to split this file into `config.d/models.yaml`, `config.d/workloads.yaml`, `config.d/rhaiis.yaml` (deploy defaults), and so on. The core framework already supports `config.d/` and it might be easier to maintain.

projects/rhaiis/orchestration/cli.py
```diff
@@ -0,0 +1,286 @@
...
+        run as run_guidellm_benchmark,
+    )
+    from projects.guidellm.toolbox.run_guidellm_benchmark.main import (
+        wait_guidellm_benchmark_task,
+    )
+    from projects.rhaiis.toolbox.deploy_kserve_isvc.main import run as deploy_kserve_isvc
+    from projects.rhaiis.toolbox.wait_isvc_ready.main import run as wait_isvc_ready
+
+    benchmark_timeout = benchmark_cfg.get("timeout", 14400)
+    wait_guidellm_benchmark_task._retry_config["attempts"] = max(1, benchmark_timeout // 10)
```
albertoperdomo2: I think we should pass the timeout to the runner (`run_guidellm_benchmark.run(timeout=...)`) and let it own its retry logic, rather than doing it here.

projects/rhaiis/orchestration/prepare_rhaiis.py
```diff
@@ -0,0 +1,58 @@
...
+    logger.info(f"Preparing namespace {ns} for rhaiis benchmarks")
+
+    result = run.run("oc whoami", capture_stdout=True, check=False)
+    if result.returncode != 0:
+        raise RuntimeError("Cannot connect to cluster")
+    logger.info(f"Connected to cluster as {result.stdout.strip()}")
+
+    result = run.run(f"oc get namespace {ns}", check=False)
+    if result.returncode != 0:
+        run.run(f"oc create namespace {ns}")
```
albertoperdomo2: Use `oc()` DSL instead of `run.run()` for `oc` commands:

```
from projects.core.dsl.utils.k8s import oc, oc_get_json
```

projects/rhaiis/orchestration/cli.py
```diff
@@ -0,0 +1,286 @@
...
+
+def _merge_env_vars(accelerator: str, model: dict) -> dict:
+    base = dict(config.project.get_config("rhaiis.env_vars") or {})
+    base.update(model.get("env_vars", {}))
+    accel_vars = config.project.get_config(f"rhaiis.accelerator_env_vars.{accelerator}") or {}
+    base.update(accel_vars)
+    return base
+
+
+def _build_guidellm_args(
```
albertoperdomo2: This hardcodes specific guidellm flags, so adding a new guidellm flag requires changing Python code instead of just adding a config key. In `llm_d` we iterate over `benchmark.get("args", {})` and converts any key/value pairs to CLI flags.

projects/rhaiis/orchestration/cli.py
albertoperdomo2, line 1: This file has no `--preset` option despite having a `presets.d/` directory. The presets are only usable via the CI path, so the CLI path needs to specify manually `--model ...` instead of `--preset ...`.

albertoperdomo2 (changes requested):
Thanks @MML-coder! Here is one extra thought: I think the RHAIIS project is missing a [phase_inputs.py](https://github.com/openshift-psap/forge/blob/main/projects/llm_d/orchestration/phase_inputs.py) file, defining contracts between phases. Right now, everything is raw dicts: config values are fetched ad-hoc with string keys (`config.project.get_config("rhaiis.deploy")`), passed as kwargs or dicts. That is fragile in the long run IMHO.

projects/rhaiis/toolbox/capture_isvc_state/main.py
MML-coder: good suggestion but I would defer to a follow-up. It requires modifying the projects/kserve/ toolbox which llm-d  projects depend on.The fact that both llmd and rhaiis are using different crd,. I would probably have to follow up later once we have end to end pipeline working.

projects/guidellm/toolbox/run_guidellm_benchmark/main.py
```diff
@@ -72,6 +72,10 @@ def validate_parameters(args, ctx):
     ctx.benchmark_name = args.name
     ctx.full_image = f"{args.image}:{args.version}"
 
+    # Scale retry attempts to match the caller's timeout
+    delay = wait_guidellm_benchmark_task._retry_config.get("delay", 10)
+    wait_guidellm_benchmark_task._retry_config["attempts"] = max(1, args.timeout // delay)
```
albertoperdomo2: I think the runner should self-configure retries to match the caller's timeout. This implementation mutates a private internal dict (`_retry_config`) on a shared function object. 

IMHO the runner should accept retry scaling through its public API. Either:
- Add an attempts parameter to `run()` and let the @retry read from ctx instead of from a hardcoded decorator argument
- Compute the attempts inside `wait_guidellm_benchmark_task` itself (e.g. read `args.timeout` and `self._retry_config["delay"]` within the task body to decide whether to keep retrying based on elapsed time)

Or something like this that avoids reaching into decorator internals. It can be done in a separate PR, but for sure this mutation shouldn't land in this one either.

projects/rhaiis/orchestration/runtime_config.py
```diff
@@ -0,0 +1,107 @@
...
+
+def merge_env_vars(accelerator: str, model: dict) -> dict:
+    base = dict(config.project.get_config("rhaiis.env_vars") or {})
+    base.update(model.get("env_vars", {}))
+    accel_vars = config.project.get_config(f"rhaiis.accelerator_env_vars.{accelerator}") or {}
+    base.update(accel_vars)
+    return base
+
+
+def build_guidellm_args(
```
albertoperdomo2, line 88: This implementation diverges a bit from the repository convention for this kind of functions. 

- It does not do `_` to `-` conversion (forcing config to use pre-hyphenated keys). Something like this would work:

```
cli_key = key.replace("_", "-")
guidellm_args.append(f"--{cli_key}={value}")
```

- It injects caller-supplied values (model, data, rate, max-seconds) into the config dict before rendering. 
- It doesn't handle list values (e.g. `rates: [300, 200, 100]` would render as Python list repr instead of `300,200,100`)

That also causes the benchmark args in [config.yaml](https://github.com/openshift-psap/forge/pull/68/changes#diff-d0b1677cc7a435ad45f0e383ac1c81591308749469f3649c4ddc6fd10dca5031) to not be as clear.

projects/rhaiis/orchestration/test_phase.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _capture_and_cleanup(deployment_name: str, namespace: str) -> None:
+    from projects.rhaiis.toolbox.capture_isvc_state.main import run as capture_isvc_state
+
+    logger.info("Capturing state")
+    try:
+        capture_isvc_state(name=deployment_name, namespace=namespace)
+    except Exception:
+        logger.warning("Capture failed, continuing with cleanup")
```
albertoperdomo2: Use `logger.warning("Capture failed, continuing with cleanup", exc_info=True)` to avoid losing stack traces.

projects/rhaiis/toolbox/wait_isvc_ready/main.py
```diff
@@ -0,0 +1,96 @@
...
+    health_result = shell.run(
+        f"oc exec {pod_name} -n {args.namespace} -c kserve-container -- "
+        f"curl -sf http://localhost:8080/health",
+        check=False,
+        log_stdout=False,
+    )
+    if health_result.returncode != 0:
+        return f"Health endpoint not responding on {pod_name} (InferenceService is Ready, proceeding)"
+
+    return f"Health check passed on pod {pod_name}"
```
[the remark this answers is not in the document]
albertoperdomo2: @MML-coder this is true too.

projects/rhaiis/orchestration/test_phase.py
```diff
@@ -0,0 +1,105 @@
+from __future__ import annotations
+
+import logging
+
+from projects.rhaiis.orchestration import runtime_config
+
+logger = logging.getLogger(__name__)
+
+
+def run(
```
albertoperdomo2, line 10: This takes 12 keyword arguments, and both `cli.py` and `test_rhaiis.py` compute all 12 identically before passing them through. Any new config field requires changes in **3** files. Consider whether it could accept fewer top-level params (e.g. model_key, workload_key, namespace) and derive the rest internally via `runtime_config`, similar to how llm_d's [test_phase.run()](https://github.com/openshift-psap/forge/blob/main/projects/llm_d/orchestration/test_phase.py#L53) reads config directly.

albertoperdomo2 (changes requested):

projects/rhaiis/orchestration/config.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+benchmarks:
+  guidellm:
+    enabled: true
+    name: guidellm-benchmark
+    image: ghcr.io/vllm-project/guidellm:v0.6.0
+    timeout: 14400
+    pvc_size: 10Gi
+    args:
+      backend: openai_http
```
albertoperdomo2: `build_guidellm_args` now correctly does `key.replace("_", "-")`, but the config keys themselves here are wrong.  For example, `backend: openai_http`  -> `--backend=openai_http`, but GuideLLM expects `--backend-type`, so it should be `backend_type: openai_http`.

MML-coder: Thans! 

verified that it works. 

`
source ~/test_foo/python3_virt/bin/activate && \
  PYTHONPATH=$PWD python3 -m projects.rhaiis.orchestration.cli test \
    --model qwen3-0_6b \
    --workload profile1 \
    --namespace kserve-e2e-perf \
    --image-pull-secret npalaska-image-pull \
    --service-account-name sa \
    --rates 1,5 \
    --max-seconds 60
`


The output generated local is:


ls -al /tmp/forge_20260612-1430/
total 32
drwxr-xr-x@  9 memehta  wheel    288 Jun 12 14:57 .
drwxrwxrwt  30 root     wheel    960 Jun 12 15:01 ..
drwxr-xr-x@  3 memehta  wheel     96 Jun 12 14:30 000__ci_metadata
drwxr-xr-x@  6 memehta  wheel    192 Jun 12 14:30 001__deploy_kserve_isvc
drwxr-xr-x@  4 memehta  wheel    128 Jun 12 14:30 002__wait_isvc_ready
drwxr-xr-x@  6 memehta  wheel    192 Jun 12 14:56 003__run_guidellm_benchmark
drwxr-xr-x@  5 memehta  wheel    160 Jun 12 14:57 004__capture_isvc_state
drwxr-xr-x@  4 memehta  wheel    128 Jun 12 14:57 005__cleanup_isvc
-rw-r--r--@  1 memehta  wheel  12456 Jun 12 14:30 config.yaml

albertoperdomo2 (approved):

projects/rhaiis/toolbox/capture_isvc_state/main.py
```diff
@@ -0,0 +1,221 @@
...
+    entrypoint,
+    execute_tasks,
+    shell,
+    task,
+)
+
+
+@entrypoint
+def run(*, name: str, namespace: str = ""):
+    return execute_tasks(locals())
```
kpouget, line 13: this is missing a docstring documentation

projects/rhaiis/toolbox/cleanup_isvc/main.py
```diff
@@ -0,0 +1,59 @@
...
+    retry,
+    task,
+)
+from projects.core.dsl.utils.k8s import oc, oc_resource_exists
+
+
+@entrypoint
+def run(*, name: str, namespace: str):
+    return execute_tasks(locals())
+
```
kpouget, line 16: same, needs a docstring documentation

projects/rhaiis/toolbox/deploy_kserve_isvc/main.py
```diff
@@ -0,0 +1,78 @@
...
+        return f"Namespace {args.namespace} exists"
+    oc("create", "namespace", args.namespace)
+    return f"Created namespace {args.namespace}"
+
+
+@task
+def render_and_apply_servingruntime(args, context):
+    output_path = args.artifact_dir / "servingruntime.yaml"
+    template.render_template_to_file("servingruntime.yaml.j2", output_path)
+    oc("apply", "-f", str(output_path))
```
kpouget, line 65: please save this in the `src` directory

projects/rhaiis/toolbox/deploy_kserve_isvc/main.py
```diff
@@ -0,0 +1,78 @@
...
+    return f"Applied ServingRuntime {args.deployment_name}"
+
+
+@task
+def render_and_apply_inferenceservice(args, context):
+    output_path = args.artifact_dir / "inferenceservice.yaml"
+    template.render_template_to_file("inferenceservice.yaml.j2", output_path)
+    oc("apply", "-f", str(output_path))
```
kpouget, line 73: please save this in the `src` directory

projects/rhaiis/toolbox/wait_isvc_ready/main.py
```diff
@@ -0,0 +1,104 @@
...
+
+    ready = condition_status(isvc, "Ready")
+    if ready == "True":
+        return f"InferenceService {args.name} is Ready"
+
+    conditions = isvc.get("status", {}).get("conditions", [])
+    reasons = [f"{c['type']}={c.get('status', '?')}({c.get('reason', '')})" for c in conditions]
+    raise RetryFailure(f"InferenceService {args.name} not ready: {', '.join(reasons)}")
+
+
```
kpouget, line 62: @MML-coder , please add a task here to capture the relevant information about the ISVC that has been deployed. Look at the llmisvc flavor.
Have it in a `@always` task, so that you capture it even when the wait fails
Capture at least:
- the isvc, the servingruntime
- the yaml of the pods, the status -owide, the logs
store all of that in the `artifacts` dir
(and have this task below `verify_health`)

projects/rhaiis/orchestration/test_rhaiis.py
```diff
@@ -0,0 +1,21 @@
...
+
+init = runtime_config.init
+
+
+@config.requires(
+    model_key="tests.rhaiis.model_key",
+    workload_key="tests.rhaiis.workload_key",
+    namespace="rhaiis.namespace",
+)
+def test(_cfg):
```
kpouget, line 16: I think you can get rid of this file

PR #72, opened 2026-06-10
feat: Add llm-d release profiles

albertoperdomo2:
# Summary

Add common `llm_d` release profiles and extend the shared GuideLLM launcher to support rate-dependent benchmark expansions.

This PR adapts the common Benchflow deployment and benchmark profiles to the current Forge `llm_d` shape instead of copying Benchflow definitions literally.

## `llm_d` profiles

- Add deployment-oriented presets:
  - `deployment-approximate-prefix-cache`
  - `deployment-precise-prefix-cache`
  - `deployment-distributed-default`
- Factor shared smoke settings into `smoke-base`
- Keep smoke presets explicit about their scheduler source while avoiding duplication

## `llm_d` benchmark workloads

- Add `heavy-heterogeneous`
- Add `multi-turn`
- Keep `concurrent-1k-1k`
- Model these as `workloads.benchmarks.*` entries in Forge config

## Shared GuideLLM launcher

- Move `build_guidellm_args(...)` into shared `guidellm` utils
- Support list-valued benchmark args such as `rates: [...]`
- Support rate-aware expansion only when a benchmark argument contains `{rate}`, `{N*rate}`, or `{rate*N}`
- Keep plain `--rates=...` benchmarks as a single GuideLLM invocation
- Run one GuideLLM benchmark per rate only for rate-dependent workloads such as:
  - `prefix_count={2*rate}`
  - `max_requests={10*rate}`

## Documentation and tests

- Update the `llm_d` README to reflect the current orchestration entrypoints and multi-rate `{rate}` expansion behavior
- Add unit coverage for:
  - rate expansion
  - job rendering behavior
  - CLI argument building
  - orchestration preset/profile validation

## Behavior notes

- Benchmark configuration resolution merges workload defaults with benchmark-specific settings
- Plain multi-rate workloads remain a single GuideLLM run
- Rate-dependent workloads expand into one run per rate and write per-run result artifacts

## Release Notes

* **New Features**
  * Added rate-based benchmark expansion for GuideLLM, including `{rate}` expression substitution and per-rate job execution.
  * GuideLLM results are now collected per expanded run with an index manifest.
  * Introduced new LLM-D deployment profiles for approximate/precise prefix caching and distributed default scaling.
  * Improved LLM-D orchestration to run per resolved run-spec (model/profile combinations) with scoped artifacts.
* **Documentation**
  * Updated LLM-D orchestration docs for run-spec workflow and rate expansion.
* **Tests**
  * Added extensive unit and end-to-end tests covering rate expansion, job rendering, and rate-split result parsing.
* **Chores**
  * Updated Caliper export configuration with a dedicated workspace setting.


projects/llm_d/orchestration/test_phase.py
```diff
@@ -11,6 +11,7 @@
 from projects.core.library.run import SignalInterrupt
 from projects.core.orchestration.utils.k8s import ensure_namespace
 from projects.guidellm.toolbox.run_guidellm_benchmark import main as run_guidellm_benchmark_command
+from projects.guidellm.toolbox.run_guidellm_benchmark.utils import build_guidellm_args
```
kpouget: can you find a way to make it clear that this is a public helper? 
like

```suggestion
from projects.guidellm.toolbox.run_guidellm_benchmark.public import build_guidellm_args
```
or `helpers` or something like that?

projects/llm_d/orchestration/config.d/workloads.yaml
```diff
@@ -31,3 +31,27 @@ benchmarks:
       rates: [300, 200, 100, 50, 1]
       data: prompt_tokens=1000,output_tokens=1000
       max_seconds: 600
+
+  heavy-heterogeneous:
+    job_name: guidellm-benchmark
+    image: ghcr.io/vllm-project/guidellm:v0.5.4
+    pvc_size: 1Gi
+    timeout_seconds: 3600
```
kpouget: can we have those in a common place, to avoid their duplication?

projects/llm_d/orchestration/config.d/deployments.yaml
```diff
@@ -0,0 +1,161 @@
+approximate-prefix-cache:
```
kpouget: I wouldn't store this in the config, more in adhoc `projects/llm_d/orchestration/manifests/deployments/name.yaml` and only the keys in the config, like:
```
approximate-prefix-cache: manifests/deployments/name.yaml
precise-prefix-cache: ...
```

kpouget: and I think there's a lot of duplication between these two files/objects, why not saving only the variable parts?

projects/llm_d/orchestration/ci.py
```diff
@@ -29,6 +29,30 @@ def init():
...
+            len(explicit_presets),
+        )
+        return
+
+    preset_name = config.project.get_config("runtime.default_preset", None)
+    if not preset_name:
+        return
+
+    logger.info("Applying fallback runtime.default_preset: %s", preset_name)
+    config.project.apply_preset(preset_name)
```
kpouget: I'm not sure to understand why you need a default preset?
at first glance I would suggest to get rid of it, but curious about how you see it

to keep your behavior, I think you can change the code to something like that:
```
explicit_presets = config.project.get_config("project.args", [], print=False) or []
if not explicit_presets:
    runtime_preset = config.project.get_config("runtime.default_preset", None)
    if runtime_preset:
        config.project.set_config("project.args", [runtime_preset])

... continue the default initialization ...
```
EDIT: not 100% sure it can help, actually

kpouget:
nice, great work @albertoperdomo2 , let's merge this 
/lgtm

PR #83, opened 2026-06-25
fournos: gitops: update the task name for proper ordering

kpouget:
## Summary by CodeRabbit

* **Chores**
  * Standardized pipeline step names across several workflows for more consistent ordering and clearer execution sequencing.
  * Updated artifact export flow so cleanup runs before exports in the full pipeline, with export behavior now tied to the cleanup result.
  * Kept pipeline tasks and parameters functionally the same otherwise.

### DIFF
--- a/fournos/gitops/base/workflows/pipeline-full.yaml
+++ b/fournos/gitops/base/workflows/pipeline-full.yaml
@@ -17,7 +17,7 @@ spec:
       type: string
 
   tasks:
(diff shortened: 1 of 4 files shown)


kpouget:
@ashtarkb FYI I'll merge this to enforce the correct ordering of the Pod and pod files
you'll tell me if that doesn't go as expected

PR #112, opened 2026-07-07
fix: leftover label in test phase mcp-gw

ashtarkb:
## Summary
- The `_cleanup_servers()` function in the MCP Gateway test phase was using a `experiment=scale-out` label selector to delete MCPServerRegistrations, HTTPRoutes, and DestinationRules between test iterations. This label was never applied to any of these resources — they are labeled with `forge.openshift.io/component=mock-mcp` (via `MOCK_MCP_LABEL`).
- As a result, cleanup between test iterations was a no-op, leaving stale resources that caused subsequent iterations to fail.

## Test plan
- [x] Verified `experiment=scale-out` is not set anywhere in the codebase
- [x] Verified `MOCK_MCP_LABEL` matches the label applied by `apply_infrastructure`
- [ ] Run `matrix-demo-1` preset with multi-iteration test to confirm cleanup works between iterations

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of Kubernetes resources during test runs by deleting the correct resource types using the appropriate label selector, ensuring related artifacts are removed consistently.
* **New Features**
  * Added a new `small-scale` orchestration preset to run smaller experiments with predefined server counts, concurrency, target, and timing parameters.

### DIFF
--- a/projects/mcp_gateway/orchestration/presets.d/presets.yaml
+++ b/projects/mcp_gateway/orchestration/presets.d/presets.yaml
@@ -45,10 +45,18 @@ matrix-demo-1:
   experiment.warmup_seconds: 5
   experiment.tools_per_server: 10
 
-
 small-scale:
   runtime.mock_server: perf-mock-server
   experiment.servers: [1, 15, 45, 60]
+  experiment.concurrency: [500]
+  experiment.target: [gateway]
+  experiment.duration_seconds: 240
+  experiment.warmup_seconds: 60
+  experiment.tools_per_server: 10
+
+large-scale:
+  runtime.mock_server: perf-mock-server
+  experiment.servers: [75, 90, 110, 130, 140]
   experiment.concurrency: [500]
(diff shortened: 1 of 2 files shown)


projects/mcp_gateway/orchestration/presets.d/presets.yaml
```diff
@@ -44,3 +44,21 @@ matrix-demo-1:
...
+  experiment.tools_per_server: 10
+
+large-scale:
+  runtime.mock_server: perf-mock-server
+  experiment.servers: [75, 90, 110, 130, 140]
+  experiment.concurrency: [500]
+  experiment.target: [gateway]
+  experiment.duration_seconds: 240
+  experiment.warmup_seconds: 60
+  experiment.tools_per_server: 10
```
kpouget, line 64: I'll fix the EOL at EOF later (there should be a GHA that prevents that)

PR #121, opened 2026-07-12
[mcp_gateway] Replace MLflow historical data with S3-based KPI pipeline

ashtarkb:
## Summary

- Replace MLflow-based historical data fetching with Caliper's standard S3 import/export pipeline for KPI comparison
- Rewrite Slack notifications to use S3-imported historical `kpis.json` files with version-aware filtering (skips same-version runs)
- Add full S3 configuration (bucket, vault, import/export settings) and enable KPI CSV generation

## Details

The previous implementation queried MLflow directly for historical run data, which had timing issues (called during `do_test` before MLflow export) and didn't align with how other projects (skeleton, llm_d) handle historical data.

This PR switches to Caliper's standard approach:
1. After KPI generation, S3 import downloads previous runs' `kpis.json` into `historical_data/`
2. The `analyse_kpis` step compares current vs historical KPIs
3. S3 export uploads the current run's KPIs for future comparisons
4. Notifications read from `historical_data/` and filter by `mcp_gateway_version` to find the most recent *different* version for meaningful comparison

## Test plan

- [ ] Verify S3 bucket `mcp-gateway-cpt-data` is created and accessible
- [ ] Run a nightly test and confirm KPIs are exported to S3
- [ ] Run a second test and confirm historical data is imported and comparison table appears in Slack
- [ ] Verify version filtering works (same version runs are skipped in comparison)

### DIFF
--- a/fournos/gitops/base/workflows/task-forge-step.yaml
+++ b/fournos/gitops/base/workflows/task-forge-step.yaml
@@ -83,7 +83,7 @@ spec:
           export KUBECONFIG=$TARGET_KUBECONFIG
           export CLUSTERLESS_MODE="false"
           echo "Target cluster info:"
-          kubectl cluster-info
+          oc cluster-info
         else
           if [[ "$KUBECONFIG_SECRET_PARAM" == "fournos-clusterless-placeholder" ]]; then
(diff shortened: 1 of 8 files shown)


kpouget:
@ashtarkb can this be closed?

PR #131, opened 2026-07-17
[llm-d] Minor improvements

kpouget:


albertoperdomo2:
Looks good to me, once the test is done, we should be good to go.

kpouget:
thanks @albertoperdomo2 , merging

PR #133, opened 2026-07-20
feat: add nightly pipeline to forge using clusterless pipeline

ashtarkb:
## Summary

Adds a clusterless nightly pipeline that automatically detects new image versions and triggers test runs when a new version is found.

**How it works:**
1. A CronJob creates a clusterless FournosJob with `pipeline: "nightly"` daily
2. The `receive-image` task calls the project's image receiver to find the latest available version (for mcp_gateway: checks ghcr.io for the latest commit with a published image)
3. The `confirm-nightly` task calls the project's verifier to get the last tested version (for mcp_gateway: queries MLflow experiment runs), compares versions, and creates a full FournosJob if a new version is detected

**Architecture:**
- Abstract interfaces (`ImageReceiver`, `NightlyVerifier`) in `projects/core/nightly/` with built-in retry
- Projects implement concrete strategies in their `resolvers/` package (auto-discovered by `NAME` attribute)
- Adding a new source = drop a new file, no other modifications needed

**Files added/modified:**
- `projects/core/nightly/` — shared interfaces and phase handlers
- `projects/mcp_gateway/resolvers/` — GHCR receiver, MLflow verifier, CronJob manifest
- `fournos/gitops/base/workflows/pipeline-nightly.yaml` — the Tekton pipeline
- `fournos/gitops/base/workflows/task-forge-step.yaml` — generalized env var export loop
- `projects/agentic_tools/ci_base.py` — registers nightly phases
- `projects/mcp_gateway/orchestration/config.yaml` — nightly config section

## Summary by CodeRabbit

* **New Features**
  * Added a nightly automation pipeline that resolves the latest image version and triggers verification jobs only when the version hasn’t been tested yet.
  * Introduced a new daily scheduled MCP Gateway nightly run.
  * Added support for `/pipeline nightly <source>` directives.
  * Enabled configurable nightly verification using registry and MLflow data.
  * Added a job submission option to control whether the launcher waits for completion.
* **Bug Fixes**
  * Improved exposure of FournosJob environment variables to forge steps.
  * Updated Service Mesh readiness handling during platform installation.


fournos/gitops/base/workflows/task-forge-step.yaml
```diff
@@ -106,10 +106,10 @@ spec:
...
         }
 
-        # Export environment variables from FournosJob spec
-        export_env_var_from_fjob PULL_BASE_SHA
-        export_env_var_from_fjob PULL_PULL_SHA
-        export_env_var_from_fjob PULL_NUMBER
+        # Export all environment variables from FournosJob spec.env
+        for var_name in $(cat "$ARTIFACT_DIR/fournos_fjob.yaml" | yq -r '.spec.env // {} | keys[]'); do
+          export_env_var_from_fjob "$var_name"
+        done
```
kpouget, line 112: I don't remember why with chose with Aleksei not to do this in Fournos 🤔
Could be interesting to go back to Fournos (PRs and doc) and see if there's any explanation

projects/agentic_tools/ci_base.py
```diff
@@ -134,7 +134,7 @@ def list_vaults(self) -> list[str]:
 
     def should_init_vaults(self, phase: str) -> bool:
         """Return False to skip vault init for certain phases."""
-        return phase != "resolve-fournos-config"
+        return phase not in ("resolve-fournos-config", "receive-image", "confirm-nightly")
```
kpouget: you can try to see if that can be stored in the config. Makes it easier to reuse the code and/or configure it

projects/agentic_tools/ci_base.py
```diff
@@ -176,6 +176,17 @@ class path like ``projects.foo.notifications.MyProvider``) and
...
 
+    _NIGHTLY_PHASES: dict[str, PhaseSpec] = {
+        "receive-image": PhaseSpec(
+            func="projects.core.nightly.receive_image:run",
+            help="Nightly: resolve latest available image version.",
+        ),
+        "confirm-nightly": PhaseSpec(
+            func="projects.core.nightly.confirm:run",
+            help="Nightly: check S3 for last run, create FournosJob if new version.",
+        ),
```
kpouget: I'm not sure to understand why you need two phases for that? 🤔

projects/core/nightly/base_receiver.py
```diff
@@ -0,0 +1,46 @@
...
+                last_error = e
+                if attempt < MAX_ATTEMPTS:
+                    logger.warning(
+                        "Receiver attempt %d/%d failed: %s. Retrying in %ds...",
+                        attempt, MAX_ATTEMPTS, e, RETRY_DELAY_SECONDS,
+                    )
+                    time.sleep(RETRY_DELAY_SECONDS)
+        raise RuntimeError(
+            f"Receiver failed after {MAX_ATTEMPTS} attempts"
+        ) from last_error
```
kpouget: I don't really like the idea of having that at framework level, with the delay and retry count hardcoded 🤔 
I think the error should be handled (and retried) as close to the issue as possible (ie, usually, in a the toolbox script)

projects/core/nightly/base_verifier.py
```diff
@@ -0,0 +1,46 @@
...
+                last_error = e
+                if attempt < MAX_ATTEMPTS:
+                    logger.warning(
+                        "Verifier attempt %d/%d failed: %s. Retrying in %ds...",
+                        attempt, MAX_ATTEMPTS, e, RETRY_DELAY_SECONDS,
+                    )
+                    time.sleep(RETRY_DELAY_SECONDS)
+        raise RuntimeError(
+            f"Verifier failed after {MAX_ATTEMPTS} attempts"
+        ) from last_error
```
kpouget: same as for the receiver, not sure this retry mechanism belongs here

projects/core/nightly/confirm.py
```diff
@@ -0,0 +1,131 @@
...
+            "Did the receive-image step run successfully?"
+        )
+
+    version = version_file.read_text().strip()
+    if not version:
+        raise RuntimeError("resolved_version.txt is empty")
+    return version
+
+
+def _create_fournos_job(project: str, version: str) -> None:
```
kpouget: there's already a toolbox script for that, in the `fournos_launcher` project
unless there's something fundamentally different it would be better to reuse it

projects/core/nightly/confirm.py
```diff
@@ -0,0 +1,131 @@
...
+    """Entrypoint for the confirm-nightly phase."""
+    logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+
+    project = config.project.get_config("project.name")
+    logger.info("Project: %s", project)
+
+    image_version = _read_resolved_version()
+    logger.info("New version available: %s", image_version)
+
+    module = importlib.import_module(f"projects.{project}.resolvers")
```
kpouget: maybe `nightly_resolvers` ? 🤔 
```suggestion
    module = importlib.import_module(f"projects.{project}.nightly_resolvers")
```

projects/core/nightly/receive_image.py
```diff
@@ -0,0 +1,42 @@
...
+    version = receiver.get_latest_version()
+    logger.info("Resolved version: %s", version)
+
+    artifact_base = Path(os.environ["ARTIFACT_BASE_DIR"])
+    version_file = artifact_base / "resolved_version.txt"
+    version_file.write_text(version)
+    logger.info("Wrote version to %s", version_file)
+
+    artifact_dir = Path(os.environ["ARTIFACT_DIR"])
+    (artifact_dir / "resolved_version.txt").write_text(version)
```
kpouget: I think `env` should be initialized when this happens

projects/mcp_gateway/nightly_resolvers/cronjob-ghcr.yaml
```diff
@@ -0,0 +1,50 @@
+# CronJob to trigger the nightly pipeline for mcp_gateway (GHCR source).
```
kpouget, line 1: I wonder where this file belongs, wrt to gitops and all 🤔 
I like the idea to have it alongside with the project, but tbd

projects/mcp_gateway/resolvers/cronjob-ghcr.yaml
```diff
@@ -0,0 +1,50 @@
...
+# Runs daily at 2:00 AM UTC. Creates a clusterless FournosJob that
+# uses the "nightly" pipeline to check for new versions on ghcr.io.
+#
+# Apply manually:
+#   oc apply -f projects/mcp_gateway/resolvers/cronjob-ghcr.yaml
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: nightly-mcp-gateway-ghcr
+  namespace: fournos-workloads
```
kpouget: `psap-automation`, no?

projects/core/nightly/confirm.py
ashtarkb: I was considering this. 
Do we want that this step of creating the fjob will be also responsible for waiting until the fournos job is completed?
I think that it better to have this as run the validation and start the job rather then waiting until the fjob that was created finishes. 
What do you think?

projects/mcp_gateway/resolvers/cronjob-ghcr.yaml
```diff
@@ -0,0 +1,50 @@
...
+  jobTemplate:
+    spec:
+      backoffLimit: 0
+      template:
+        spec:
+          serviceAccountName: fournos
+          restartPolicy: Never
+          containers:
+            - name: submit
+              image: bitnami/kubectl:latest
```
kpouget: please never use short image names 😛  [dockerism]
please never use docker.io images 😛 [docker.io rate-limit hits us hard sometimes]

There should be a quay.io version of this (or with `oc-cli`)

but why do we need to *launch* the `FournosJob`, can't the CronJob do it? 🤔

projects/core/nightly/confirm.py
kpouget: https://github.com/openshift-psap/forge/blob/main/projects/fournos_launcher/toolbox/submit_and_wait/main.py

I see that there's no `wait=True | wait=False` flag. I considered it at some point but I did it another way.

you can add a flag `wait: true` by default, and set it to `wait: false` in your launcher. Use the `EarlyReturn` object to interrupt the script before `wait_for_job_completion` (use the `@when` annotation if that works)

in the case of the nightly we indeed don't want to wait 👍🏻

projects/mcp_gateway/nightly_resolvers/cronjob-ghcr.yaml
ashtarkb: I think:
1. let's currently keep them at a project level for simplicity, validation and initial testing to make sure that everything is working correctly. 
2. Later I think as more and more projects are going to be onboarded and we will have more schedules we will need to add a gitops operations to take the cronjob file, validate it and deploy it on the mgmt cluster automatically. we will need to also implement a more sophisticated scheduling mechanism that will be able to observe what cron jobs are already there, what clusters are used and understand how can we manage this because I think  that if I'm creating a cronjob to run nightly at say 00:00 and another team mate want's to create another nightly in 00:00 we need to adjust (Also based on the length of the test), I know that we have queue but maybe we can find a more efficient  way to handle this such as "this cluster is already scheduled, let's run on another cluster that is currently free" and in this way we receive better usability of the clusters. but all of this in another PR's and later work.
What do you think?

projects/agentic_tools/ci_base.py
ashtarkb: I want to separate this 2 tasks in the pipeline to receive better visibility when something failed.
If the confirm failed but we actually had a new version this is another concern then "there was a timeout on the first step" and should be handled differently.

kpouget: the logic makes sense, 
but I wonder if we can't have it simpler at the pipeline level (one step)
and handle the "visibility" aspect at the notification level. Something like:
```
try:
  receive_image()
except:
  add_notification_file("receive_image failed", level=FATAL)
  return
try:
  confirm()
except:
  add_notification_file("confirm failed", level=FATAL)
```
(`add_notification_file` already exists, but it doesn't accept the level flag)

projects/core/nightly/handler.py
```diff
@@ -0,0 +1,117 @@
...
+def _create_fournos_job(project: str, version: str) -> None:
+    """Create a FournosJob for the full test run using the fournos_launcher toolbox."""
+    from projects.fournos_launcher.toolbox.submit_and_wait.main import run as submit_and_wait
+
+    namespace = os.environ.get("FOURNOS_WORKLOAD_NAMESPACE", "fournos-workloads")
+    cluster = config.project.get_config("nightly.cluster")
+    version_key = config.project.get_config("nightly.version_key")
+    pipeline = config.project.get_config("nightly.pipeline")
+    preset = config.project.get_config("nightly.preset", "")
+    owner = config.project.get_config("nightly.owner", "nightly-pipeline")
```
kpouget: can you use the `@requires` config annotation?
(that itf isn't perfect but better than nothing)

projects/fournos_launcher/toolbox/submit_and_wait/main.py
```diff
@@ -208,6 +210,9 @@ def submit_fournos_job(args, ctx):
     # Apply the job manifest (will raise CalledProcessError with full details on failure)
     shell.run(f"oc apply -f {ctx.manifest_file}")
 
+    if not args.wait:
+        return EarlyReturn(f"Submitted FournosJob: {ctx.final_job_name} (wait=False)")
```
kpouget, line 214: just a note, in the future we may want to check that the jobs goes into resolve then pending before continuing

projects/mcp_gateway/nightly_resolvers/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+            if issubclass(attr, NightlyVerifier) and attr is not NightlyVerifier:
+                name = getattr(attr, "NAME", None)
+                if name:
+                    _verifiers[name] = attr
+
+
+def get_receiver() -> ImageReceiver:
+    """Return the image receiver based on NIGHTLY_SOURCE env var."""
+    _discover()
+    source = os.environ.get("NIGHTLY_SOURCE", "ghcr")
```
kpouget: you should have access to the config at this point,
so better store the source in the config,
and use a `configOverride` to set its value, if necessary

projects/mcp_gateway/nightly_resolvers/ghcr_receiver.py
```diff
@@ -0,0 +1,95 @@
...
+
+from projects.core.nightly.base_receiver import ImageReceiver
+
+logger = logging.getLogger(__name__)
+
+REPO = "Kuadrant/mcp-gateway"
+IMAGE = "kuadrant/mcp-gateway"
+COMMITS_URL = f"https://api.github.com/repos/{REPO}/commits?per_page=20"
+TOKEN_URL = f"https://ghcr.io/token?service=ghcr.io&scope=repository:{IMAGE}:pull"
+MANIFEST_URL = f"https://ghcr.io/v2/{IMAGE}/manifests/"
```
kpouget: REPO and IMAGE should be in the config

projects/mcp_gateway/nightly_resolvers/ghcr_receiver.py
```diff
@@ -0,0 +1,95 @@
...
+                if attempt < MAX_ATTEMPTS:
+                    logger.warning(
+                        "GHCR attempt %d/%d failed: %s. Retrying in %ds...",
+                        attempt,
+                        MAX_ATTEMPTS,
+                        e,
+                        RETRY_DELAY_SECONDS,
+                    )
+                    time.sleep(RETRY_DELAY_SECONDS)
+        raise RuntimeError(f"GHCR receiver failed after {MAX_ATTEMPTS} attempts") from last_error
```
kpouget, line 64: could you try to rewrite what with a toolbox script ? makes it easier to troubleshoot

kpouget: (the full file, I think)

projects/mcp_gateway/nightly_resolvers/mlflow_verifier.py
```diff
@@ -0,0 +1,136 @@
...
+                        MAX_ATTEMPTS,
+                        e,
+                        RETRY_DELAY_SECONDS,
+                    )
+                    time.sleep(RETRY_DELAY_SECONDS)
+        raise RuntimeError(f"MLflow verifier failed after {MAX_ATTEMPTS} attempts") from last_error
+
+    def _load_connection(self) -> dict:
+        secrets_dir = Path(os.environ.get("FOURNOS_SECRETS", "/var/run/secrets/fournos"))
+        secret_path = secrets_dir / self.secret_name / self.secret_file
```
kpouget: you don't need that, the vault should give you the full path directly

projects/mcp_gateway/nightly_resolvers/mlflow_verifier.py
```diff
@@ -0,0 +1,136 @@
+"""MLflow nightly verifier for mcp_gateway.
```
kpouget: could be interesting to have this in the `caliper.orchestration` package for reuse

projects/mcp_gateway/nightly_resolvers/mlflow_verifier.py
```diff
@@ -0,0 +1,136 @@
...
+    NAME = "mlflow"
+
+    def __init__(self):
+        from projects.core.library import config
+
+        self.experiment = config.project.get_config(
+            "caliper.export.backend.mlflow.config.experiment"
+        )
+        self.workspace = config.project.get_config("caliper.export.backend.mlflow.config.workspace")
+        vault_cfg = config.project.get_config("caliper.export.backend.mlflow.secrets.vault")
```
kpouget: if moved to caliper.orchestration, please use the @requires config annotation

projects/mcp_gateway/nightly_resolvers/ghcr_receiver.py
ashtarkb: I think that it makes less sense in this context. 
If I want to create for one project 3 image receivers, I dont want to explode the toolboxes dir and I want to have the abstraction to have one engine to receive any image receiver and use it.
I think that having the ImageReceiver as the abc class and then subcallses implements it makes more sense. This is what I was thinking about from the beginning, each project will be able to run nightly for upstream, downstream, anything, with a single engine.

kpouget: > I dont want to explode the toolboxes dir

this part is not an issue, with the right `artifact_dirname_suffix` it's easy to distinguish the toolbox directories
but fine if you prefer to have it in Python 👍🏻

ashtarkb:
@kpouget let me know what do you think.
The pipelines works correctly and I think that most of your comments/concerns were addressed. 
The failed job above is not related to the pipeline but to all of the operators that were installed in the smoke testing cluster. will be further tested in another PR, the nightly pipeline works correctly.

projects/fournos_launcher/toolbox/submit_and_wait/main.py
kpouget: FYI, I've done this "future" step in my PR

projects/caliper/orchestration/mlflow_verifier.py
```diff
@@ -0,0 +1,137 @@
...
+                    )
+                    time.sleep(RETRY_DELAY_SECONDS)
+        raise RuntimeError(f"MLflow verifier failed after {MAX_ATTEMPTS} attempts") from last_error
+
+    def _load_connection(self) -> dict:
+        secret_path = vault.get_vault_content_path(self.secret_name, self.secret_file)
+
+        if secret_path is None or not secret_path.exists():
+            logger.warning("MLflow secret not found in vault, assuming no previous run")
+            return {}
```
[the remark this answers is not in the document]
kpouget: @ashtarkb I agree with that, better raise an exception when this happens

kpouget:
```
INFO: get_config: project.name --> mcp_gateway
INFO: get_config: nightly.source --> ghcr
INFO: get_config: nightly.sources.ghcr --> {'receiver': 'projects.mcp_gateway.nightly_resolvers.ghcr_receiver.GHCRReceiver', 'repo': 'Kuadrant/mcp-gateway', 'image': 'kuadrant/mcp-gateway', 'mlflow_experiment': 'mcp-gw-cpt-ghcr'}
INFO: Project: mcp_gateway, source config: {'receiver': 'projects.mcp_gateway.nightly_resolvers.ghcr_receiver.GHCRReceiver', 'repo': 'Kuadrant/mcp-gateway', 'image': 'kuadrant/mcp-gateway', 'mlflow_experiment': 'mcp-gw-cpt-ghcr'}
INFO: get_config: nightly.sources.ghcr --> {'receiver': 'projects.mcp_gateway.nightly_resolvers.ghcr_receiver.GHCRReceiver', 'repo': 'Kuadrant/mcp-gateway', 'image': 'kuadrant/mcp-gateway', 'mlflow_experiment': 'mcp-gw-cpt-ghcr'}
INFO: Using receiver: GHCRReceiver
INFO: Resolved version: f7e2b44c0ff2406f5fac41f3501b952f7545bc7e
INFO: get_config: nightly.verifier --> projects.caliper.orchestration.mlflow_verifier.CaliperMLflowVerifier
INFO: get_config: caliper.export.backend.mlflow.config.experiment --> mcp-gateway-cpt
INFO: get_config: caliper.export.backend.mlflow.config.workspace --> mcp-gw-cpt
INFO: get_config: caliper.export.backend.mlflow.secrets.vault --> {'name': 'psap-forge-mlflow-export', 'mlflow_secret': 'mlflow-secret.yaml'}
INFO: get_config: nightly.mlflow_verifier.version_pattern --> -v(.+?)-\d{8}-\d{6}$
INFO: get_config: nightly.mlflow_verifier.strip_prefix --> sha-
INFO: Using verifier: CaliperMLflowVerifier
INFO: Latest MLflow run: forge-mcp-gateway-20260720-163611
WARNING: Could not extract version from run name: forge-mcp-gateway-20260720-163611
INFO: Last tested version: '<none>'
INFO: New version detected: f7e2b44c0ff2406f5fac41f3501b952f7545bc7e (previous: <none>)
```
(can you mute the `InsecureRequestWarning`? should already been done in other parts of Caliper)

does [this](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_forge/133/pull-ci-openshift-psap-forge-main-fournos/2079882745340760064/artifacts/fournos/fournos/artifacts/task_logs/forge-mcp-gateway-20260722-110318-00-nightly-pod.log) look good?

would be nice to have an example with this correctly set
```
INFO: Last tested version: '<none>'
```

(and I deleted the unrelated jobs launches/errors to keep the last nightly launch nearby)

projects/mcp_gateway/orchestration/pr_args.py
```diff
@@ -77,4 +79,14 @@ def parse_project_directives(comment_text: str) -> tuple[dict[str, Any], list[st
             parsed_directives.append(line)
             logger.info("Parsed mcp_gateway version: %s", parts[0])
 
+        # Parse /pipeline nightly <source> — extract source from pipeline directive
+        if line.startswith("/pipeline nightly "):
```
kpouget, line 85: interesting idea to overload the `/pipeline` directive :D 

now it appears [twice](https://github.com/openshift-psap/forge/pull/133#issuecomment-5045029346) in the notification message, maybe you can find a way to avoid saving it here?

projects/caliper/orchestration/mlflow_verifier.py
```diff
@@ -0,0 +1,137 @@
...
+        self.workspace = _cfg.workspace
+        self.secret_name = _cfg.vault_cfg["name"]
+        self.secret_file = _cfg.vault_cfg["mlflow_secret"]
+        self._version_re = re.compile(_cfg.version_pattern)
+        self._strip_prefix = _cfg.strip_prefix or ""
+
+    def get_last_tested_version(self) -> str:
+        connection = self._load_connection()
+        if not connection:
+            return ""
```
kpouget: this should be a ValueError I think

projects/caliper/orchestration/mlflow_verifier.py
kpouget: this should be a FileNotFound or ValueError, the vault config must be correct

projects/caliper/orchestration/mlflow_verifier.py
```diff
@@ -0,0 +1,137 @@
...
+
+        data = load_mlflow_secrets_yaml(secret_path)
+        validate_mlflow_secrets(data)
+        return data
+
+    def _query_latest_run(self) -> str:
+        import mlflow
+
+        if self.workspace:
+            os.environ["MLFLOW_WORKSPACE"] = self.workspace
```
kpouget: can you unset this in a `finally` handler?

projects/caliper/orchestration/mlflow_verifier.py
```diff
@@ -0,0 +1,137 @@
...
+
+        client = mlflow.tracking.MlflowClient()
+
+        experiment = client.get_experiment_by_name(self.experiment)
+        if experiment is None:
+            logger.warning("Experiment '%s' not found in MLflow", self.experiment)
+            return ""
+
```
kpouget: shouldn't be just a warning

we need to think about the bootstrap (avoid failing when creating a new nightly test), but that shouldn't be handled in `caliper` but in the project itself

projects/caliper/orchestration/mlflow_verifier.py
```diff
@@ -0,0 +1,137 @@
...
+        runs = client.search_runs(
+            experiment_ids=[experiment.experiment_id],
+            order_by=["start_time DESC"],
+            max_results=1,
+        )
+
+        if not runs:
+            logger.info("No runs found in experiment '%s'", self.experiment)
+            return ""
```
kpouget: same as above, should be an exception, not a warning

projects/caliper/orchestration/mlflow_verifier.py
```diff
@@ -0,0 +1,137 @@
...
+        run_name = latest_run.info.run_name or ""
+        logger.info("Latest MLflow run: %s", run_name)
+
+        return self._extract_version(run_name)
+
+    def _extract_version(self, run_name: str) -> str:
+        match = self._version_re.search(run_name)
+        if not match:
+            logger.warning("Could not extract version from run name: %s", run_name)
+            return ""
```
kpouget: should be a ValueError

projects/core/nightly/base_verifier.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class NightlyVerifier(ABC):
+    """Checks historical run data to determine if a version was already tested."""
+
+    @abstractmethod
+    def get_last_tested_version(self) -> str:
+        """Return the version string of the last successful run.
+
+        Returns empty string if no previous run exists.
```
kpouget: will return a version string or raise exceptions

projects/core/nightly/handler.py
```diff
@@ -0,0 +1,139 @@
...
+    logger.info("Using verifier: %s", type(verifier).__name__)
+    last_version = verifier.get_last_tested_version()
+    logger.info("Last tested version: '%s'", last_version or "<none>")
+
+    if image_version == last_version:
+        logger.info("Version %s already tested. No new run needed.", image_version)
+        ci_lib.add_notification_file(
+            "nightly-noop", f"Version {image_version} already tested — no-op."
+        )
+        (env.ARTIFACT_DIR / "result.txt").write_text("NO-OP")
```
kpouget, line 76: Can you add a similar file when a version is detected and a job is launched.
Include the name of the job and the version

projects/core/nightly/handler.py
```diff
@@ -0,0 +1,139 @@
...
+        return
+
+    logger.info("New version detected: %s (previous: %s)", image_version, last_version or "<none>")
+    _create_fournos_job(project, image_version, source_cfg)
```
kpouget, line 80: I'm not sure about the `wait=False` flag. Wouldn't it be easier for tracking if the launcher waits ?

projects/core/nightly/handler.py
kpouget: ah, I see below that the notification file is created, but it's not taken into account in the GH notification coming from `submit.py` ...

but the file wasn't created though
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift-psap_forge/133/pull-ci-openshift-psap-forge-main-fournos/2079882745340760064/artifacts/fournos/fournos/artifacts/000__ci_metadata/notifications/

(I'll fix the vault warning tomorrow)

kpouget: ah, no, forget, I was mixing things up.

What's happening is that you don't have an artifacts upload step in the nightly pipeline, so the logs don't end up in MLFlow [and I was confused by that, as the artifact export also sends the GH notification]

kpouget:
great work Avi, let's see that live !
/approve
/lgtm

PR #156, opened 2026-08-03
[mcp_gateway] Reduce the log verbosity

kpouget:


kpouget:
@ashtarkb PTAL, this should make the run.log file and task.log files easier to read

PR #173, opened 2026-08-17
[rhaiis] Add missing Model Furnace models and new NVFP4/Qwen3.x/DeepSeek models

ssaketh-ch:
## Summary
- Add 15 models from Model Furnace that were missing from Forge, with TP
  sizes and vllm args matching gitlab.cee.redhat.com/psap/model-furnace.
- Add 14 new models not yet in either system: NVFP4 quantizations,
  Qwen3.6 and Qwen3.8 families, and DeepSeek-V4-Flash.
- Fix two broken preset aliases in presets.yaml that pointed at undefined
  model keys (phi-4 and llama-405b).

## Motivation
Model Furnace and Forge had diverged on which models each tracks. This
closes the gap for the rhaiis project and adds recently released models
on top.

## Testing
Models added from Model Furnace have been validated there. New models
(NVFP4, Qwen3.6, Qwen3.8, DeepSeek-V4-Flash) will need a benchmark run
to confirm the TP values and vllm args are correct.

### From Model Furnace (not yet in Forge)

| Key | HF Model ID | TP |
|-----|-------------|----|
| `phi-4` | `microsoft/phi-4` | 1 |
| `phi-4-w4a16` | `RedHatAI/phi-4-quantized.w4a16` | 1 |
| `phi-4-w8a8` | `RedHatAI/phi-4-quantized.w8a8` | 1 |
| `phi-4-fp8` | `RedHatAI/phi-4-FP8-dynamic` | 1 |
| `llama-3-1-8b-fp8` | `RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8-dynamic` | 1 |
| `llama-3-1-8b-w8a8` | `RedHatAI/Meta-Llama-3.1-8B-Instruct-quantized.w8a8` | 1 |
| `llama-3-1-8b-w4a16` | `RedHatAI/Meta-Llama-3.1-8B-Instruct-quantized.w4a16` | 1 |
| `llama-3-1-405b` | `meta-llama/Llama-3.1-405B-Instruct` | 8 |
| `llama-3-1-405b-fp8` | `RedHatAI/Meta-Llama-3.1-405B-Instruct-FP8-dynamic` | 8 |
| `llama-3-1-405b-w8a8` | `RedHatAI/Meta-Llama-3.1-405B-Instruct-quantized.w8a8` | 8 |
| `qwen3-5-35b-fp8` | `RedHatAI/Qwen3.5-35B-A3B-FP8-dynamic` | 1 |
[...]

PR #181, opened 2026-08-20
[rhaiis] Add missing workload profiles 5-7 and composite benchmark presets

hp2419:
## Summary
- Sync workload profiles from model-furnace that were missing in forge:
  - **profile5**: 100k/1k ultra-long context (rates [1,2,5], 10 samples)
  - **profile6**: multi-turn with prefix caching (5 turns, prefix_tokens=512)
  - **profile7**: heavy heterogeneous (prompt 50-30k, output 20-8k)
- Add workload key preset mappings for profiles 5, 6, 7
- Add composite benchmark presets for common test matrices:
  - **benchmark-standard**: profile1 + profile2
  - **benchmark-long-context**: profile4 + profile5
  - **benchmark-full-sweep**: all profiles 1-7
  - **benchmark-multi-turn**: profile6
  - **benchmark-heterogeneous**: profile7

## Test plan
- [ ] Verify profile5/6/7 workload definitions match model-furnace
- [ ] Verify composite presets resolve correctly via runtime_config
- [ ] Run ci-test preset to confirm no regressions

### DIFF
--- a/projects/rhaiis/orchestration/presets.d/benchmarks.yaml
+++ b/projects/rhaiis/orchestration/presets.d/benchmarks.yaml
@@ -6,3 +6,39 @@ benchmark:
   rhaiis.profiler.enabled: true
   caliper.postprocess.csv_dashboard.enabled: true
   rhaiis.agent_analysis.enabled: false
+
+benchmark-standard:
+  extends:
+  - benchmark
+  tests.rhaiis.workload_key:
+  - profile1
+  - profile2
+
+benchmark-long-context:
+  extends:
+  - benchmark
+  tests.rhaiis.workload_key:
+  - profile4
+  - profile5
+
+benchmark-full-sweep:
+  extends:
+  - benchmark
+  tests.rhaiis.workload_key:
+  - profile1
+  - profile2
+  - profile3
+  - profile4
+  - profile5
+  - profile6
+  - profile7
+
+benchmark-multi-turn:
+  extends:
+  - benchmark
+  tests.rhaiis.workload_key: profile6
+
+benchmark-heterogeneous:
+  extends:
+  - benchmark
+  tests.rhaiis.workload_key: profile7


hp2419:
cc @Harshith-umesh @albertoperdomo2 — could you review this when you get a chance?

projects/rhaiis/orchestration/presets.d/benchmarks.yaml
```diff
@@ -6,3 +6,39 @@ benchmark:
...
+
+benchmark-multi-turn:
+  extends:
+  - benchmark
+  tests.rhaiis.workload_key: profile6
+
+benchmark-heterogeneous:
+  extends:
+  - benchmark
+  tests.rhaiis.workload_key: profile7
```
kpouget, line 44: random question from a rhaiis-outsider: is there no way to give a meaningful name to the profiles?
`profile6` seems to be `multi-turn` and `profile7`  is `heterogeneous`,

but are these number profiles mandatory? 🤔

hp2419: Great question! The numbered names (profile1–7) are inherited from [model-furnace](https://gitlab.cee.redhat.com/psap/model-furnace), where they're used in regression comparison (`PROFILE_MAP`), payload pattern matching
   (`_PROFILE_PATTERNS`), and benchmark launch scripts. Renaming them in forge would break the alignment between the two systems and make cross-system result comparisons harder.

   That said, the `preset layer` is exactly where we add meaningful names. The workload keys stay as `profile1`–`profile7`, but users interact through presets like `benchmark-multi-turn`,
   `benchmark-heterogeneous`, `benchmark-long-context`, etc. 

   We could also consider adding aliases in `workloads.yaml` (e.g. `multi-turn` pointing to the same config as `profile6`) if we want meaningful names at the workload level too, without breaking the numbered convention.

Harshith-umesh:
@hp2419  I'm closing this PR since @sacherukuri has already added this functionality in this PR: https://github.com/openshift-psap/forge/pull/173 a few days ago.

PR #206, opened 2026-09-07
docs(security): prohibit literal secret args

ashtarkb:
Prevent secret values from appearing in command text that may be captured in artifacts or logs.

## Summary by CodeRabbit

- **Documentation**
  - Added security guidance prohibiting secrets from being included directly in command arguments or command text.
  - Documented using securely created temporary files and file-based references instead.
  - Added examples illustrating safe and unsafe secret-handling patterns.


kpouget:
thanks @ashtarkb 

/lgtm

PR #209, opened 2026-09-08
[rhaiis] Add SGLang presets for simplified test configuration

aas008:
## Summary

Add an `sglang-ci-quick` preset for quick SGLang CI testing, and split the monolithic `presets.yaml` into separate files by category.

### New preset: `sglang-ci-quick`

Lightweight SGLang validation test — eliminates the need for multiple `/var` overrides:
- Engine: SGLang
- Model: Qwen3-0.6B (small, fast)
- Workload: ci-quick (rates and max_seconds come from `config.d/workloads.yaml`)
- Warmup/profiler/analysis: disabled (fast test)

### Split `presets.d/presets.yaml` into separate files

Per review feedback, split the monolithic presets file into:
- **`engines.yaml`** — accelerator (nvidia/amd), engine (vllm/sglang/trtllm), and feature toggles (prefix-cache-on/off)
- **`ci_tests.yaml`** — ci-test, ci-quick, and sglang-ci-quick presets
- **`workloads.yaml`** — profile1-7 workload key mappings
- **`models.yaml`** — all model alias presets

### Fixes
- Removed `workloads.ci-quick.rates` and `workloads.profile1.max_seconds` overrides from `sglang-ci-quick` — these are already defined in `config.d/workloads.yaml` and are picked up automatically via `workload_key`
- Added singular `tests.rhaiis.workload_key: ci-quick` alongside the plural `workload_keys` — `get_test_workload_key()` reads the singular key and raises `KeyError` if absent

### Before (8 /var overrides needed)
```
/test fournos rhaiis nvidia
/var rhaiis.engine: sglang
/var tests.rhaiis.model_key: qwen3-0_6b
/var tests.rhaiis.workload_keys: ["ci-quick"]
/var workloads.ci-quick.rates: [1, 50, 100]
/var tests.rhaiis.warmup: false
/var rhaiis.profiler.enabled: false
/var caliper.postprocess.csv_dashboard.enabled: true
/var rhaiis.agent_analysis.enabled: false
```

### After (single preset + cluster)
```
/test fournos rhaiis nvidia sglang-ci-quick hera
/cluster hera
```

Settings can still be overridden with `/var` as needed.


aas008:
@Harshith-umesh , this PR is to avoid the long commands we leave as comments for simplified sglang tests.

projects/rhaiis/orchestration/presets.d/sglang.yaml
```diff
@@ -0,0 +1,17 @@
+__multiple: true
+
+# SGLang quick CI test - lightweight test with small model
+sglang-ci-quick:
```
Harshith-umesh: why not add this in this file: https://github.com/openshift-psap/forge/blob/d5ab3bfc3f337bd4d50d5837d47b8d0ce13c9914/projects/rhaiis/orchestration/presets.d/presets.yaml#L35

projects/rhaiis/orchestration/presets.d/sglang.yaml
```diff
@@ -0,0 +1,17 @@
...
+  extends:
+    - sglang
+  tests.rhaiis.model_key: qwen3-0_6b
+  tests.rhaiis.workload_keys: ["ci-quick"]
+  workloads.ci-quick.rates: [1, 50, 100]
+  workloads.profile1.max_seconds: 30
+  tests.rhaiis.warmup: false
+  rhaiis.profiler.enabled: false
+  caliper.postprocess.csv_dashboard.enabled: true
```
Harshith-umesh: dashboard should be disabled

projects/rhaiis/orchestration/presets.d/sglang.yaml
aas008: Yeah that makes more sense! I have moved it there.

kpouget: @aas008 @Harshith-umesh , `presets.d/presets.yaml` is already large, you should split it in like
- `presets.d/model.yaml` (maybe multiple of them) 
- `presets.d/workload.yaml`
- `presets.d/ci_tests.yaml`
- `presets.d/engine.yaml`

(just as examples)

projects/rhaiis/orchestration/presets.d/presets.yaml
```diff
@@ -41,6 +41,21 @@ ci-quick:
   tests.rhaiis.slack_notify_always: true
   rhaiis.agent_analysis.enabled: false
 
+sglang-ci-quick:
+  extends:
+    - sglang
+  tests.rhaiis.model_key: qwen3-0_6b
+  tests.rhaiis.workload_keys: ["ci-quick"]
+  workloads.ci-quick.rates: [1, 50, 100]
```
Harshith-umesh, line 49: this and the next line are already defined in ci-quick

projects/rhaiis/orchestration/presets.d/presets.yaml
```diff
@@ -41,6 +41,21 @@ ci-quick:
...
+  workloads.profile1.max_seconds: 30
+  tests.rhaiis.warmup: false
+  rhaiis.profiler.enabled: false
+  caliper.postprocess.csv_dashboard.enabled: false
+  rhaiis.agent_analysis.enabled: false
+  rhaiis.cluster_tag: "hera2"
```
Harshith-umesh: this assumes Hera is always going to be used for the ci test, this and the next 2 lines can be removed. they're already in the Hera or zeus preset. We are using preset chaining in the GitHub comment.

projects/rhaiis/orchestration/presets.d/sglang.yaml
Harshith-umesh: makes sense Kevin! @aas008 could you please do this?

projects/rhaiis/orchestration/presets.d/presets.yaml
kpouget: >   `rhaiis.cluster_tag: "hera2"`

@Harshith-umesh this configuration shouldn't be necessary.
The cluster name can be accessed at the config key `ci_job.cluster` (see the bottom of [this file](https://mlflow.apps.psap-automation.ibm.rhperfscale.org/#/experiments/264/runs/d23f20c30cde40b4aed754a0356fb9fb/artifacts/03__test/run.log?workspace=forge-rhaiis))

And since a recent PR, you can define presets that are applied based on the cluster name.
Like in this llm-d [preset file](https://github.com/openshift-psap/forge/blob/main/projects/llm_d/orchestration/presets.d/cluster_config.yaml). The presets need to start with `
cluster_NAME` to be automatically applied
```
INFO: apply_presets_from_cluster_config: no preset named 'cluster_hera', nothing to apply.
```

But here you'll have to find a trick to rename `hera` into `hera2`.

Harshith-umesh: @kpouget this config is required for the dashboard.

kpouget: no, this value is required. Question is, where do you take the value from?
currently, you take it from the config file, and you populate it via a preset.

My suggestion is to take it from the config ... but from the field populated from the FournosJob specification

Harshith-umesh: I'm a little confused here Kevin, but when we specify the cluster preset we're already passing this cluster_tag config: https://github.com/openshift-psap/forge/blob/d5ab3bfc3f337bd4d50d5837d47b8d0ce13c9914/projects/rhaiis/orchestration/presets.d/clusters.yaml#L3

kpouget: there are two things
```
  rhaiis.cluster_tag: "hera2"
```
1. you set this value, because you need to consume at some point. You need to know which cluster the test ran on. What I'm trying to explain is that Forge has _another_ way to know it. If you read the config key `ci_job.cluster` instead of `rhaiis.cluster_tag`, you'll have the same information (except that it's `hera` vs `hera2`, this will need a bit of a hack)

2. the [hera preset](https://github.com/openshift-psap/forge/blob/d5ab3bfc3f337bd4d50d5837d47b8d0ce13c9914/projects/rhaiis/orchestration/presets.d/clusters.yaml#L3). Currently you need to specify 1/ the cluster you want to use `fjob.spec.cluster: hera` and the 2/ preset to apply `fjob.spec.engine.executionEngine.forge.args: [hera, ....]`. If you rename your preset to `cluster_hera`, the preset will be automatically loaded during the initialization, and 2/ won't be necessary anymore

Harshith-umesh: so the cluster tag is not straightforward to know from the cluster name, it's a coincidence that the tags are hera2 and zeus2 but for the old-zeus cluster there is no cluster tag. For Mi300x cluster the tag is just "2". so we would need an explicit cluster tag.

projects/rhaiis/orchestration/presets.d/presets.yaml
aas008:
```
workloads.ci-quick.rates: [1, 50, 100]
workloads.profile1.max_seconds: 30
```

I don't see them in ci-quick.

Harshith-umesh: https://github.com/openshift-psap/forge/blob/ce696a3957dc7a883a72f9e55fe4a7059665b791/projects/rhaiis/orchestration/config.d/workloads.yaml#L1

projects/rhaiis/orchestration/presets.d/sglang.yaml
aas008: @kpouget @Harshith-umesh - I have split presets.yaml into 4 files as per the thread. Please take a look.

projects/rhaiis/orchestration/presets.d/presets.yaml
```diff
@@ -41,6 +41,19 @@ ci-quick:
...
   rhaiis.agent_analysis.enabled: false
 
+sglang-ci-quick:
+  extends:
+    - sglang
+  tests.rhaiis.model_key: qwen3-0_6b
+  tests.rhaiis.workload_keys: ["ci-quick"]
+  tests.rhaiis.warmup: false
+  rhaiis.profiler.enabled: false
+  caliper.postprocess.csv_dashboard.enabled: false
```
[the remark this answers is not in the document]
aas008: This is intentional. users can override if they need. 
cc: @Harshith-umesh

projects/rhaiis/orchestration/presets.d/presets.yaml
aas008: I've removed the cluster tag var @Harshith-umesh , user must specify the cluster for the test I think.

projects/rhaiis/orchestration/presets.d/ci_tests.yaml
```diff
@@ -0,0 +1,33 @@
...
+  caliper.postprocess.csv_dashboard.enabled: false
+  tests.rhaiis.slack_notify_always: true
+  rhaiis.agent_analysis.enabled: false
+
+sglang-ci-quick:
+  extends:
+    - sglang
+  tests.rhaiis.model_key: qwen3-0_6b
+  tests.rhaiis.workload_key: ci-quick
+  tests.rhaiis.workload_keys: ["ci-quick"]
```
Harshith-umesh, line 28: @aas008 you need the version as well in the preset

aas008: Do you mean the vllm image version or version as in the label for dashboard?

aas008: Added tests.rhaiis.version: "sglang-ci-quick-deleteme"

PR #210, opened 2026-09-08
[rhaiis] Add ConfigIQ benchmark profile

Rehan164:
Added:
- Add the ConfigIQ 1000/1000 workload profile
- Sweep concurrency at 1, 2, 5, 10, 25, 50, 75, 100, 200, and 300
- Limit each concurrency run to 120 seconds
- Exclude the first 15 seconds of each concurrency run as GuideLLM warmup
- Retain the existing one-time 60-second deployment warmup
- Add the ConfigIQ preset and documentation

This is Tier 1 of a 2 Tier sweep for configiq:
1. Run a dense concurrency grid of 1, 2, 5, 10, 25, 50, 75, 100, 200, and 300. Each point runs for 120 seconds, with the first 15 seconds excluded as warmup.
2. Adaptive second pass: A per-run profile that adds extra points around the knee point once the first pass reveals the curve shape

## Summary by CodeRabbit

* **New Features**
  * Added the `configiq` workload preset for running sweeps with 1,000 prompt and output tokens.
  * Configured rates from 1–300 and a maximum runtime of 450 seconds.
  * The preset uses the standard one-time deployment warmup.

* **Documentation**
  * Updated workload profile documentation to include `configiq`.
  * Removed per-rate warmup details and the GuideLLM 15-second exclusion note.


Harshith-umesh:
@Rehan164 looks good, could you trigger a quick test for the new profile that we're adding. reduce the time of each run to 30s with warmup of 15s for a quick ci test.

projects/rhaiis/orchestration/config.d/workloads.yaml
```diff
@@ -40,6 +40,12 @@ profile7:
   rates: [1, 50, 100, 200, 300]
   max_seconds: 450
 
+configiq:
+  data: "prompt_tokens=1000,output_tokens=1000"
+  rates: [1, 2, 5, 10, 25, 50, 75, 100, 200, 300]
+  max_seconds: 120
```
Harshith-umesh, line 46: this should be 135s, if you want 120s of the run + 15s warmup for each concurrency

Rehan164: Set back to 450 for now

PR #216, opened 2026-09-14
[rhaiis] preserve MLflow link for early failures

Harshith-umesh:
## Summary

- Persist the pre-created MLflow destination before deployment in a job-level marker.
- Reuse that marker for MLflow artifact export and Slack URL construction when readiness fails before benchmark labels exist.
- Keep the job-level marker ahead of benchmark child metadata and add focused tests.

## Testing

- `ruff check` passed for changed implementation and tests.
- `ruff format --check` passed.
- `python3 -m py_compile` passed for changed Python files.
- Focused pytest collection is blocked in this environment because `jsonpath_ng` is not installed.
- The repository container test path is unavailable because Podman is not connected.

## Summary by CodeRabbit

* **New Features**
  * MLflow destinations can now be persisted at the job level and reused across related benchmark and test runs.
  * Preparation, preflight, and test workflows automatically create or reuse the job-level destination when running in Fournos CI.
  * Job-level destination metadata takes precedence over benchmark metadata, with shared artifact-root discovery supported.

* **Bug Fixes**
  * Invalid or incomplete MLflow destinations now produce clear errors instead of being silently ignored.
  * MLflow setup failures now propagate reliably, preventing workflows from continuing with missing destination information.


projects/caliper/orchestration/export.py
```diff
@@ -380,6 +334,91 @@ def precreate_mlflow_run_if_configured(_cfg, force=False) -> dict[str, str] | No
...
+    try:
+        temporary_fd, temporary_name = tempfile.mkstemp(
+            prefix=f".{marker_path.name}.", dir=marker_path.parent
+        )
+        temporary_path = Path(temporary_name)
+        with os.fdopen(temporary_fd, "w", encoding="utf-8") as marker_file:
+            yaml.safe_dump(normalized, marker_file, sort_keys=False)
+            marker_file.flush()
+            os.fsync(marker_file.fileno())
+
```
kpouget: I'm not sure why we need a primitive open here 🤔

kpouget:
```
        # Hard-linking the completed temporary file publishes it atomically and
        # fails without overwriting a marker created by another writer.
```
there's no concurrency in forge for the time being.
```if exists: raise exception``` will be enough to preserve any race condition

File: projects/llm_d/orchestration/test_phase.py [hunk 2]
```diff
@@ -356,14 +319,26 @@ def run_all_tests(stop_on_error: bool = False) -> int:
...
-    for run_spec in runtime_config.get_run_specs():
+    for run_spec in run_specs:
         with runtime_config.activate_run_spec(run_spec):
             with env.NextArtifactDir(run_spec.artifact_dirname):
                 try:
-                    exit_code = do_test()
+                    exit_code = do_test(
+                        mlflow_destination=mlflow_destination,
+                        precreate_mlflow=False,
+                    )
```
kpouget: why do we have to carry this down to the test ??

I think this should remain untouched

the job level marker should be created once and for all, then each test creates its own destination

projects/llm_d/orchestration/test_phase.py
```diff
@@ -470,13 +450,15 @@ def do_test() -> int:
...
+            precreate_mlflow_run_if_configured,
+            write_mlflow_destination_marker,
+        )
 
         mlflow_destination = precreate_mlflow_run_if_configured()
-    except Exception:
-        logger.error("MLflow run pre-creation failed; continuing", exc_info=True)
-        mlflow_destination = None
+        if mlflow_destination:
+            write_mlflow_destination_marker(mlflow_destination)
```
kpouget: this isn't necessary, it's done below for each of the tests

```
        create_test_labels(mlflow_destination=mlflow_destination)
```

kpouget:
thinking about it @Harshith-umesh , this should be done in the entrypoint of the CI commands (except the resolve).
When the prepare/preflight/test steps start, they should check if the file exists at ARTIFACT_BASE_DIR (name should be improved...)

so that we know where we send for any step, before any failure

kpouget:
```
mlflow_destination:
  run_id: ec2a5fdbeb2a40e3bb6f03277ec31813
  experiment_id: '264'
  workspace: forge-rhaiis
```
@Harshith-umesh can you make sure that the mlflow dest isn't created for each of the tests? I got confused as some point but that's not necessary. Or at least that would require using the nested run capability

projects/caliper/orchestration/export.py
```diff
@@ -380,6 +333,96 @@ def precreate_mlflow_run_if_configured(_cfg, force=False) -> dict[str, str] | No
...
+        raise ValueError(f"Invalid MLflow destination in marker: {source}")
+
+    run_id = destination.get("run_id")
+    experiment_id = destination.get("experiment_id")
+    if not run_id:
+        raise ValueError(f"Incomplete MLflow destination in marker: {source}")
+    if require_experiment_id and not experiment_id:
+        raise ValueError(f"Incomplete MLflow destination in marker: {source}")
+
+    return {key: str(value) for key, value in destination.items() if value is not None}
```
kpouget: @Harshith-umesh can we return a dataclass object here ?

projects/caliper/orchestration/export.py
```diff
@@ -436,30 +479,61 @@ def precreate_mlflow_run(
     return meta
 
 
+def _read_mlflow_destinations(
+    artifact_dir: Path, *, primary_only: bool = False
+) -> list[dict[str, str]]:
+    """Read job and benchmark MLflow destinations, with optional job-only lookup."""
```
kpouget: I think we need to update all of that to remove the `optional job-only lookup` part

projects/caliper/orchestration/export.py
```diff
@@ -436,30 +479,61 @@ def precreate_mlflow_run(
...
-            run_id = dest.get("run_id", "")
-            if run_id:
-                return run_id, dest.get("experiment_id", "")
-        except (OSError, yaml.YAMLError) as e:
-            logger.warning("Failed to read test labels %s: %s", labels_file, e)
+
+    destinations = _read_mlflow_destinations(artifact_dir, primary_only=True)
+    if destinations:
+        destination = destinations[0]
+        return destination["run_id"], destination.get("experiment_id", "")
```
kpouget: this function is counter-intuitive
```
def _read_mlflow_ids_from_test_labels() -> tuple[str, str]:
```
but
```
destinations = _read_mlflow_destinations(artifact_dir, primary_only=True)
```
I think `from_test_labels` has to go away

projects/caliper/orchestration/export.py
```diff
@@ -522,208 +598,15 @@ def build_mlflow_run_url(
...
-        except (OSError, yaml.YAMLError) as e:
-            logger.warning("Failed to read test labels %s: %s", labels_file, e)
+    """Find a pre-created MLflow run ID from job or benchmark metadata."""
+    destinations = _read_mlflow_destinations(from_path, primary_only=True)
+    if destinations:
+        run_id = destinations[0]["run_id"]
+        logger.info("Found pre-created MLflow run_id: %s", run_id)
+        return run_id
 
     return None
```
kpouget: looks redundant with the function above.
Will be easier when we only have one `mlflow_destination`

projects/llm_d/orchestration/ci.py
```diff
@@ -102,6 +102,11 @@ def main(ctx, preset):
 
     init_vaults_for_phase(ctx.invoked_subcommand)
 
+    if ctx.invoked_subcommand in {"prepare", "preflight", "test"}:
+        from projects.caliper.orchestration.export import ensure_mlflow_destination_marker
```
kpouget: import can be at the top

projects/llm_d/orchestration/test_phase.py
```diff
@@ -356,14 +319,22 @@ def run_all_tests(stop_on_error: bool = False) -> int:
     Returns:
         Maximum exit code from all tests
     """
+    from projects.caliper.orchestration.export import (
+        ensure_mlflow_destination_marker,
+        read_mlflow_destination_marker,
+    )
     from projects.llm_d.orchestration import runtime_config
 
+    ensure_mlflow_destination_marker()
```
kpouget: redundant with the code in `main()`

projects/llm_d/orchestration/test_phase.py
```diff
@@ -356,14 +319,22 @@ def run_all_tests(stop_on_error: bool = False) -> int:
...
+    run_specs = runtime_config.get_run_specs()
+
     max_exit_code = 0
-    for run_spec in runtime_config.get_run_specs():
+    for run_spec in run_specs:
         with runtime_config.activate_run_spec(run_spec):
             with env.NextArtifactDir(run_spec.artifact_dirname):
                 try:
-                    exit_code = do_test()
+                    exit_code = do_test(mlflow_destination=mlflow_destination)
```
kpouget: these changes will go away with a single destination

projects/llm_d/orchestration/test_phase.py
```diff
@@ -385,13 +356,20 @@ def run_all_tests(stop_on_error: bool = False) -> int:
 def run() -> int:
     """Main test function that wraps do_test() with outcome postprocessing."""
 
+    from projects.caliper.orchestration.export import (
+        ensure_mlflow_destination_marker,
+        read_mlflow_destination_marker,
+    )
+
+    ensure_mlflow_destination_marker()
```
kpouget: redundant

projects/llm_d/orchestration/test_phase.py
```diff
@@ -452,7 +430,8 @@ def _run_finalizer(
     return primary_exc, finalizer_exc
 
 
-def do_test() -> int:
+def do_test(*, mlflow_destination: dict[str, str] | None = None) -> int:
```
kpouget: `mlflow_destination` will go away

projects/rhaiis/orchestration/test_phase.py
```diff
@@ -40,6 +39,10 @@ def run(
     namespace: str,
     deployment_name: str | None = None,
 ) -> int:
+    from projects.caliper.orchestration.export import ensure_mlflow_destination_marker
+
+    ensure_mlflow_destination_marker()
+
```
kpouget: redundant

projects/rhaiis/orchestration/ci.py
```diff
@@ -129,6 +129,11 @@ def main(ctx):
     if ctx.invoked_subcommand != "resolve-fournos-config":
         vault.init(runtime_config.get_vaults())
 
+        if ctx.invoked_subcommand in {"prepare", "preflight", "test"}:
+            from projects.caliper.orchestration.export import ensure_mlflow_destination_marker
+
+            ensure_mlflow_destination_marker()
```
kpouget: I think I suggested to rewrite this in a previous comment
```
    if ctx.invoked_subcommand != "resolve-fournos-config":
```
this is enough. Make that a == comparison, and return early if true. Same in llm-d

projects/caliper/orchestration/export.py
Harshith-umesh: done

projects/caliper/orchestration/export.py
Harshith-umesh: done

projects/caliper/orchestration/export.py
Harshith-umesh: done

projects/llm_d/orchestration/ci.py
Harshith-umesh: done

projects/llm_d/orchestration/test_phase.py
Harshith-umesh: done

projects/llm_d/orchestration/test_phase.py
Harshith-umesh: done

projects/llm_d/orchestration/test_phase.py
Harshith-umesh: done

projects/llm_d/orchestration/test_phase.py
Harshith-umesh: done

projects/rhaiis/orchestration/test_phase.py
Harshith-umesh: done

projects/rhaiis/orchestration/ci.py
Harshith-umesh: done

projects/caliper/orchestration/export.py
```diff
@@ -18,7 +18,11 @@
 import yaml
 from pydantic import ValidationError
 
-from projects.caliper.engine.constants import LEGACY_METADATA_FILE, METADATA_FILE
+from projects.caliper.engine.constants import (
+    LEGACY_METADATA_FILE,
+    METADATA_FILE,
+    MLFLOW_DESTINATION_FILE,
+)
```
kpouget, line 25: these constants should be exposed in a `public`  package, we'll do that in another PR

kpouget: same for that:
```
from projects.caliper.engine.kpi.dataclasses import MlflowDestination
```

projects/caliper/orchestration/export.py
```diff
@@ -359,42 +369,137 @@ def precreate_mlflow_run_if_configured(_cfg, force=False) -> dict[str, str] | No
...
+    if not env.running_inside_fournos():
+        logger.info("Not running inside FOURNOS CI, skipping MLflow destination marker")
         return None
 
-    return {
-        "run_id": meta["run_id"],
-        "experiment_id": meta.get("experiment_id", ""),
-        "workspace": _cfg.workspace or "",
-    }
+    normalized = _normalize_mlflow_destination(destination, source="pre-created MLflow run")
```
kpouget: we don't want to normalize, we want one way to invoke, with `destination` as `MlflowDestination`, not as `dict[str, str]`.
that's the kind of thing I've been heavily rewriting lately from AI-generated code

kpouget: the dict->destination code is used below after reading from yaml.
This can go along with the dataclass object definition

projects/llm_d/orchestration/ci.py
```diff
@@ -89,6 +90,9 @@ def main(ctx, preset):
 
     init_vaults_for_phase(ctx.invoked_subcommand)
 
+    if ctx.invoked_subcommand in {"prepare", "preflight", "test"}:
+        ensure_mlflow_destination_marker()
```
kpouget: @Harshith-umesh you don't follow nor answer my comments on this topic!
the return early is good, but we don't want this check `ctx.invoked_subcommand in {"prepare", "preflight", "test"}`

projects/llm_d/orchestration/test_phase.py
```diff
@@ -217,9 +216,7 @@ def create_test_labels(
         env.ARTIFACT_DIR,
         labels,
         kpi_labels=kpi_labels if kpi_labels else None,
-        mlflow_destination=MlflowDestination.from_dict(mlflow_destination)
-        if mlflow_destination
-        else None,
+        mlflow_destination=mlflow_destination,
```
kpouget: @Harshith-umesh  this is wrong, 
```
mlflow_destination = read_mlflow_destination_marker()
```
this is a single destination, but with this
```
write_test_labels(...mlflow_destination...)
```
you duplicate it to every tests, whereas each test is uploaded to a dedicated nested run

projects/llm_d/orchestration/test_phase.py
```diff
@@ -358,8 +355,10 @@ def run_all_tests(stop_on_error: bool = False) -> int:
     """
     from projects.llm_d.orchestration import runtime_config
 
+    run_specs = runtime_config.get_run_specs()
+
     max_exit_code = 0
-    for run_spec in runtime_config.get_run_specs():
+    for run_spec in run_specs:
         with runtime_config.activate_run_spec(run_spec):
```
kpouget: irrelevant change

projects/rhaiis/orchestration/ci.py
```diff
@@ -126,8 +127,13 @@ def main(ctx):
...
     test_rhaiis.init()
 
-    if ctx.invoked_subcommand != "resolve-fournos-config":
-        vault.init(runtime_config.get_vaults())
+    if ctx.invoked_subcommand == "resolve-fournos-config":
+        return
+
+    vault.init(runtime_config.get_vaults())
+
+    if ctx.invoked_subcommand in {"prepare", "preflight", "test"}:
```
kpouget: remove this test, we want to `ensure_mlflow_destination_marker` for any step name

projects/rhaiis/orchestration/test_phase.py
```diff
@@ -522,12 +509,16 @@ def _create_test_labels(
         "run_uuid": run_uuid,
     }
 
+    mlflow_destination = None
+    if env.running_inside_fournos():
+        from projects.caliper.orchestration.export import read_mlflow_destination_marker
+
+        mlflow_destination = read_mlflow_destination_marker()
+
```
kpouget: remove this, see llm-d comment. Same for the hunk below

projects/caliper/orchestration/export.py
```diff
@@ -359,42 +369,137 @@ def precreate_mlflow_run_if_configured(_cfg, force=False) -> dict[str, str] | No
...
+    marker_path = _mlflow_destination_path()
+    if marker_path.exists():
+        read_mlflow_destination_marker(artifact_root=marker_path.parent)
+        logger.info("Using existing MLflow destination marker: %s", marker_path)
+        return marker_path
+
+    destination = precreate_mlflow_run_if_configured()
+    if destination is None:
+        return None
+    return write_mlflow_destination_marker(destination, artifact_root=marker_path.parent)
```
kpouget: I think you need to remove `artifact_root` everywhere, it doesn't make sense:
```
marker_path = _mlflow_destination_path()
```
here you don't have it, so you use the default
```
        read_mlflow_destination_marker(artifact_root=marker_path.parent)
...
    return write_mlflow_destination_marker(destination, artifact_root=marker_path.parent)
```
here you pass back the same information, so you end up using the default flag

Harshith-umesh: Done, removed `artifact_root` from the marker APIs and use the default marker path.

projects/caliper/orchestration/export.py
Harshith-umesh: Done, YAML is converted directly to `MlflowDestination` in the reader.

projects/caliper/orchestration/export.py
Harshith-umesh: Agreed, I left the constants here and will handle exposing them in `public` in a separate PR.

projects/rhaiis/orchestration/ci.py
Harshith-umesh: Done, RHAIIS now ensures the marker for every step name.

projects/llm_d/orchestration/test_phase.py
Harshith-umesh: Removed the unrelated change.

projects/rhaiis/orchestration/test_phase.py
Harshith-umesh: Done, removed marker propagation from the RHAIIS test labels as well.

projects/llm_d/orchestration/test_phase.py
Harshith-umesh: Done, removed the job destination from per-test labels; each test keeps its own nested run.

projects/caliper/orchestration/export.py
Harshith-umesh: Agreed, removed the normalization path and kept the marker API dataclass-only.

projects/caliper/orchestration/export.py
Harshith-umesh: Same for this import, I left it unchanged for this PR.

projects/llm_d/orchestration/ci.py
Harshith-umesh: Done, marker creation now runs for every non-resolve command.

kpouget:
thanks for the good work @Harshith-umesh,
that's an important feature, I'm glad we could get it in 🙏🏻 
(and sorry for the early/previous misunderstanding with the caliper_test_metadata file, at some point I got the wrong picture and pushed you the wrong way. My bad 😅)

/lgmt

kpouget:
> /lgmt

I meant that 😅 
/lgtm

PR #229, opened 2026-09-16
feat: support GuideLLM 0.7.x CLI in benchmark job rendering

aas008:
## Summary

GuideLLM 0.7.0 removed the `benchmark` subcommand and replaced it with `guidellm run` using a new registry-based option syntax (`--backend kind=…`, `--profile kind=…`, `--constraint kind=…`, `--output kind=…`). This PR migrates FORGE's benchmark job rendering to the v0.7.x CLI format exclusively — v0.6.x support has been removed.

### CLI translation (utils.py)
- Emits `guidellm run --backend=kind=openai_http,target=… …` format for all jobs
- Covers single-run, multi-run (rate expressions), and shared-volume (sidecar) job templates
- Data spec conversion: detects source type (synthetic, file, HuggingFace) and emits `kind=…` format
- Decimal rate/rampup support: preserves `float` values for rates and `rampup_duration`
- Converts legacy `--processor=<model>` to `--tokenizer=kind=huggingface_auto,model=<model>` (v0.7.x replacement)
- Injects `--warmup` and `--rampup` into `--profile` spec (including no-rate profiles like throughput/sweep)

### Output JSON parsing (dashboard.py)
- v0.7.x moved data/token config from `payload.args.data` to `payload.config.spec.data[]` and `request_type` to `config.spec.backend.request_format`
- Parser falls back to `config.spec` paths when the `args` key is absent
- Dashboard CSV correctly populated with prompt/output token counts, turns, prefix tokens, and request type from v0.7.x output

Refs: vllm-project/guidellm v0.7.0 migration guide

## Usage

GuideLLM v0.7.x image is required. Pass the image via `/var` in the PR test comment:

```
/test fournos rhaiis nvidia ci-quick hera
/cluster hera
/var benchmarks.guidellm.image: "ghcr.io/vllm-project/guidellm:v0.7.4"
```

## Test plan

- [x] CI-quick test passed on Hera with v0.7.3: `nvidia ci-quick hera` completed in 4m 12s
- [x] Full profile test passed on Hera with v0.7.4: profiles 1, 2, 4 all completed successfully
- [x] Post-processing pipeline fully successful: parse → artifacts_to_kpis → kpis_to_mlflow → dashboard_csv
- [x] Dashboard CSV correctly extracts metrics from v0.7.x JSON output format
- [x] 26 unit tests covering: CLI arg translation, data spec conversion, rate expansion, dashboard fallbacks, processor→tokenizer conversion, rampup in no-rate profiles
[...]

PR #198, opened 2026-09-02
[llm_d] Use benchconf config repo

albertoperdomo2:
## Summary by CodeRabbit

- **New Features**
  - Benchmark runs can now use reusable external YAML configurations through `benchconf` references.
  - GuideLLM jobs accept a local configuration file, validate it, and apply it during execution.
  - Added configurable `benchconf` support, including optional custom versions.

- **Improvements**
  - Updated concurrent and heavy heterogeneous workload presets to use reusable configurations.
  - Improved handling and validation of missing or invalid configuration references.
  - Preserved inline configuration support for workloads requiring rate-based settings.

PR #234, opened 2026-09-17
rhaiis: make B200 GuideLLM jobs SCC-compatible

Harshith-umesh:
## Summary
- remove the B200-only `fsGroup: 0` preset
- add a B200-only `benchmarks.guidellm.host_users: false` preset
- render `hostUsers` conditionally so other clusters are unchanged
- cover both PVC and shared-volume GuideLLM Job templates

## Why
B200's restricted-v3 SCC rejects the generated GuideLLM Job because it contains `fsGroup: 0` and omits `hostUsers`, which must be explicitly false on that cluster.

## Validation
- `python3 -m pytest -q projects/guidellm/tests/test_run_guidellm_benchmark.py`
- 9 tests passed
- `git diff --check` passed
- Ruff still reports one unrelated pre-existing `UP038` warning in `projects/rhaiis/orchestration/test_phase.py`

## Summary by CodeRabbit

* **New Features**
  * Added an optional `hostUsers` setting for GuideLLM benchmark jobs in both PVC and shared-volume modes.
  * The setting can explicitly enable or disable host user namespaces, or remain unspecified by default.
  * Benchmark, warmup, and profiler launches now honor the configured value.
  * Updated the B200 preset to disable host users for compatibility with its restricted security configuration.

* **Documentation**
  * Documented the new GuideLLM `host_users` configuration option.


kpouget:
this will conflict with 
https://github.com/openshift-psap/forge/pull/229 from @aas008  and https://github.com/openshift-psap/forge/pull/198 from @albertoperdomo2 
I'll let you sync on the merge order of each PR

projects/rhaiis/orchestration/presets.d/clusters.yaml
```diff
@@ -24,4 +24,4 @@ b200:
   rhaiis.cluster_tag: "b200"
   rhaiis.gpu_types.nvidia: "b200"
   rhaiis.deploy.image_pull_secrets: ["npalaska-image-pull"]
-  benchmarks.guidellm.fs_group: 0
+  benchmarks.guidellm.host_users: false
```
Harshith-umesh, line 27: here @kpouget

albertoperdomo2:
I think mine can be the last one to merged, since it has the most changes in our project and I would like to make sure everything works.

kpouget:
I discussed with @Harshith-umesh offline and maybe this PR can be turned into a 1-line change, <> `pod.spec.hostUsers: false` if we can confirm that this works well on any cluster 🤞🏻

Harshith-umesh:
not required

PR #235, opened 2026-09-17
guidellm: set hostUsers false for benchmark pods

Harshith-umesh:
## Summary
- set `spec.template.spec.hostUsers: false` in both GuideLLM Job templates
- remove the B200-only `benchmarks.guidellm.fs_group: 0` preset
- add rendering tests and document the common setting

## Why
The B200 cluster's restricted-v3 SCC rejects GuideLLM benchmark Pods when `hostUsers` is omitted. This PR tests Kevin's simpler proposal: use `hostUsers: false` for all GuideLLM benchmark Pods. GuideLLM does not require the host user namespace.

## Validation
- `python3 -m pytest -q projects/guidellm/tests/test_run_guidellm_benchmark.py`
- 9 tests passed
- `git diff --check` passed

Please validate the generated Job on the existing IBM Cloud clusters before merging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated GuideLLM benchmark Jobs to use isolated host user namespaces, improving compatibility with restricted OpenShift security configurations.
  * Removed the B200-specific filesystem group override from the benchmark configuration.


Harshith-umesh:
not required

